PR #3, opened 2018-12-19
Add !inputs to explain uncontrolled vs controlled inputs

dylangarcia:
https://goshakkk.name/controlled-vs-uncontrolled-inputs-react/


BTMPL:
If we're going to have commands to explain some mechanics / topics, maybe a better idea is to have an `!explain forms`, `!explain setState` ?

dylangarcia:
@BTMPL would it be worth the effort creating extra helpers to parse 2 word commands, or should the trigger words just literally be `!explain xyz`?

PR #11, opened 2019-02-12
[WIP] transfer static message commands to json

bpas247:
I'm still in the process of transferring all of the messages over, will update the title of this branch when it's ready for review and merge.


bpas247:
So I've moved over as many static messages as I could to json, the remaining static messages are a bit long and wouldn't look very good in the json file (e.g. it would just be one long string on a single line). Any suggestions for a workaround?

vcarl:
I'd say it's alright to have some long lines in the JSON, especially since they're already long lines in the JS 🤷‍♂️ Could manually split it up into a multiline string, but not sure if that'd lead to weird breaks in the discord message.

BTMPL:
Could see if we can inject `\n` and have it parse properly. I'll pull the branch on the bot and experiment.

PR #12, opened 2019-07-07
enhance autoban to act more like a spam filter

evertbouw:


evertbouw:
doesn't look like this is going anywhere

PR #17, opened 2019-07-16
Add deduper feature to the bot

### LINKED ISSUE #15: Dedupe cross-posts
We recommend against cross-posting, it would be great to have it enforced by a bot. If somebody posts exactly the same message in more than 1 channel within a narrow time frame, Reactibot should remove one of the messages.

rafaelalmeidatk:
If a message is duplicated in one of the channels listed, the bot will delete the message and send a DM to the user. I need opinions on the time to forgive the cross-post (currently 2 minutes) and the DM message (that I literally copied from the `jobs` module hehe). Currently, the bot only reacts to the `#general`, `#need-help-0` and `#need-help-1` channels.

Closes #15

### DIFF
--- a/features/deduper.js
+++ b/features/deduper.js
@@ -1,0 +1,45 @@
+const deduper = {
+  monitoredChannels: [
+    "600025060379721760",
+    "600037597955489797",
+    "600037610005463050"
+  ],
+
+  // dupeExpireTime
+
+  lastMessageOfMember: {},
+
+  deleteMsgAndWarnUser: (msg, lastMsg) => {
+    msg.delete();
+    // prettier-ignore
+    msg.author
+        .send(`Hello there! It looks like you just posted a message to the #${msg.channel.name} channel on our server.
+			
+We don't allow cross posting, so I had to delete your message because it is idential as your last one posted on #${lastMsg.channel.name}. You can avoid this warning deleting the first message.
+
+Thank you :)
+
+:robot: This message was sent by a bot, please do not respond to it - in case of additional questions / issues, please contact one of our mods!`);
+  },
+
+  handleMessage: ({ msg, user }) => {
+    if (deduper.monitoredChannels.includes(msg.channel.id)) {
+      const userId = user.id;
+      const lastMessage = deduper.lastMessageOfMember[userId];
+      if (
+        lastMessage &&
+        !lastMessage.deleted &&
+        lastMessage.channel.id !== msg.channel.id &&
+        lastMessage.content === msg.content
+      ) {
+        deduper.deleteMsgAndWarnUser(msg, lastMessage);
+      } else {
+        deduper.lastMessageOfMember[userId] = msg;
+      }
+    }
+  }
+};
+
+module.exports = {
+  default: deduper
+};
--- a/features/deduper.js
+++ b/features/deduper.js
@@ -5,7 +5,7 @@ const deduper = {
     "600037610005463050"
   ],
 
-  // dupeExpireTime
+  dupeExpireTime: 2 * 60 * 1000, // 2 minutes to reject the duplicate as cross-post (ms)
 
   lastMessageOfMember: {},
 
@@ -15,7 +15,7 @@ const deduper = {
     msg.author
         .send(`Hello there! It looks like you just posted a message to the #${msg.channel.name} channel on our server.
 			
-We don't allow cross posting, so I had to delete your message because it is idential as your last one posted on #${lastMsg.channel.name}. You can avoid this warning deleting the first message.
+We don't allow cross-posting, so I had to delete your message because it is identical as your last one posted on #${lastMsg.channel.name}. You can avoid this warning deleting the first message or waiting 2 minutes.
 
 Thank you :)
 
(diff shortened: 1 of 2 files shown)


features/deduper.js
```diff
@@ -0,0 +1,61 @@
...
+    "565213527673929729" // #need-help-1
+  ],
+
+  dupeExpireTime: 2 * 60 * 1000, // 2 minutes to reject the duplicate as cross-post (ms)
+  cleanerInterval: 10 * 60 * 1000, // Frees up the memory from 10 to 10 minutes (ms)
+
+  lastMessageOfMember: {},
+
+  deleteMsgAndWarnUser: (msg, lastMsg) => {
+    msg.delete();
```
vcarl: Feeling like we should delete the last message, so if they cross post only the latest survives. feel like that'd be a little more subtle, so the folks who ignore their DMs and quickly spam don't get wise to what's happening.

```suggestion
    lastMsg.delete();
```

features/deduper.js
```diff
@@ -0,0 +1,61 @@
...
+        lastMessage &&
+        !lastMessage.deleted &&
+        lastMessage.channel.id !== msg.channel.id &&
+        lastMessage.content === msg.content &&
+        msg.createdTimestamp - lastMessage.createdTimestamp <
+          deduper.dupeExpireTime
+      ) {
+        deduper.deleteMsgAndWarnUser(msg, lastMessage);
+      } else {
+        deduper.lastMessageOfMember[userId] = msg;
```
vcarl: I like that this doesn't accumulate tons of messages, but it breaks in the case if they send a second message before cross posting. Maybe not super common, but would be nice to track user + channel, I think.

features/deduper.js
rafaelalmeidatk: Wouldn't it be more confusing to automatically delete the last message without the user knowing (since he is not looking at the other channel by the time it is deleted)?

features/deduper.js
rafaelalmeidatk: Yeah currently it only sees the last message. The problem with user + channel is that the user could have sent a msg "X" in a channel and 10 days later, without sending other msg in this channel but in others, he could send "X" again in a different channel and it would be considerate as a cross-post.

features/deduper.js
vcarl: Exactly. It's less visible, which in this case I think is a benefit. They're notified by the DM, not by the immediacy of seeing their message disappear.

rafaelalmeidatk: But isn't it kinda unexpected? One user may have already answered the question then the message suddenly disappears because the poster sent it on another channel, idk 🤔

vcarl: If a handful of minutes that feels like an uncommon case 🤷‍♂ I'm okay with unexpected things happening when people behave badly

features/deduper.js
BTMPL: Wouldn't the
```
 msg.createdTimestamp - lastMessage.createdTimestamp <
          deduper.dupeExpireTime
```
above prevent that?

features/deduper.js
rafaelalmeidatk: Done! Please check the bot message, I am not sure it is clear enough or I wrote something wrong.

features/deduper.js
rafaelalmeidatk: Ah, right. I was thinking about another comment when I wrote this reply.

bpas247:
What's preventing this from being merged? Any current issues with the implementation?

BTMPL:
> What's preventing this from being merged? Any current issues with the implementation?

Hi  there. Honestly, I've been away for  some time and didn't have time to look into the PRs. I'll check this one and the new one in the next 48h.

I remember we recently merged something that broke the commands but I believe it was reverted.

vcarl:
Yep just got some new motion on it, intending to review it in the next few days.

bpas247:
> I remember we recently merged something that broke the commands but I believe it was reverted.

I'm guessing you're referring to my PR? 👀

TrevorSayre:
@rafaelalmeidatk taking a look at these to see if I can help get this through

TrevorSayre:
This will also need to post to the channel `#mod-log` with the message:
> Warned ${user} for crossposting

features/deduper.js
```diff
@@ -0,0 +1,94 @@
...
+   * }
+   */
+  lastMessagesOfMember: {},
+
+  deleteMsgAndWarnUser: (msg, lastMsg) => {
+    lastMsg.delete();
+    // prettier-ignore
+    msg.author
+        .send(`Hello there! It looks like you just posted a message to the #${msg.channel.name} channel on our server.
+
```
TrevorSayre: Is this whitespace intentional?

bpas247: I would imagine it's there to add some spacing between each section of text.

rafaelalmeidatk: Yeah, similar to the message of `#jobs` channel

TrevorSayre: You specifically need three tabs? It can't be just a newline?

rafaelalmeidatk: Oh I didn't even notice it lol
Thanks for pointing out, I removed it

vcarl:
It's been 5 years :pensive: closing this. The core idea is solid but I think the exact behavior is a bit of a bikeshed problem — seems to be tricky to get consensus on how to respond

PR #18, opened 2019-10-14
Add loop messages feature

### LINKED ISSUE #13: Add configurable periodic messages
It would be great to be able to have Reactibot post messages to several channels every so often. A daily post in #jobs reminding people of the rules would be great, and a welcome post to #introductions pointing people to the guidelines and tips. 

We can iron out exact messages later, being able to set up what channel, how often, and the message would be great.

rafaelalmeidatk:
With this feature, the bot will keep a message as the last one in a channel, with an interval between each post, to keep important information more visible to users.

I swear I couldn't find a better name for the module 😅 Suggestions are welcome

Closes #13

### DIFF
--- a/features/jobs.js
+++ b/features/jobs.js
@@ -3,6 +3,8 @@ const cooldown = require("./cooldown").default;
 const jobs = {
   tags: ["forhire", "for hire", "hiring", "remote", "local"],
   handleMessage: ({ msg, user }) => {
+    if (msg.author.bot) return;
+
     let hasTags = false;
     jobs.tags.forEach(tag => {
       if (msg.content.toLowerCase().includes(`[${tag}]`)) hasTags = true;
--- a/features/loopMessages/index.js
+++ b/features/loopMessages/index.js
@@ -1,0 +1,75 @@
+/**
+ * ------------------------------
+ * Loop Messages
+ * ------------------------------
+ * The bot will be sending a determined set of messages to some channels to repeat
+ * important information, but without spamming these channels
+ *
+ * How it works:
+ * At startup and repeatedly by a given interval, for each loop message, the bot will:
+ * 1. Get the target channel of the loop message
+ * 2. Fetch the last 50 messages of the channel
+ * 3. If the last message is already the message we want to loop, do nothing (we don't want to spam the channel)
+ * 4. Delete all loop messages (so the channel isn't filled with bot messages)
+ * 5. Send the loop message
+ */
+
+const loopMessages = require("./messages.js");
+
+// time between each check to send a new loop message
+const INTERVAL_TIME = 5 * 60 * 1000; // 5 minutes (ms)
+
+function getContent(loopMessage) {
+  return loopMessage.trim();
+}
+
+function isMessageFromBot(bot, message) {
+  return message.author.id === bot.user.id;
+}
+
+function isLoopMessage(messageToCheck, loopMessage, bot) {
+  return (
+    isMessageFromBot(bot, messageToCheck) &&
+    messageToCheck.content.trim() === getContent(loopMessage.content)
+  );
+}
+
+async function sendLoopMessage(client, loopMessage) {
+  const channel = client.channels.get(loopMessage.channelId);
+  const channelMessages = await channel.fetchMessages({ limit: 50 });
+
+  // if the last message in the channel is from the bot, we don't need to send it again
+  if (isLoopMessage(channelMessages.first(), loopMessage, client)) {
+    return;
+  }
+
+  // search through all the last messages for the loop message, so we can delete it
+  channelMessages.forEach(message => {
+    if (isLoopMessage(message, loopMessage, client)) {
+      message.delete();
+    }
+  });
+
+  // now we can send the loop message in the channel
+  channel.send(getContent(loopMessage.content));
+}
+
+module.exports = {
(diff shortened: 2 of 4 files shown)


features/loopMessages/messages.js
```diff
@@ -0,0 +1,12 @@
+module.exports = [
```
rafaelalmeidatk: I am using a JS file instead of JSON because of the comments that are useful with `channelId` and the template strings for the multi-line messages

features/loopMessages/index.js
```diff
@@ -0,0 +1,75 @@
...
+        loopMessages.forEach(message => {
+          sendLoopMessage(client, message);
+        });
+      };
+
+      setInterval(() => {
+        sendLoopMessages();
+      }, INTERVAL_TIME);
+
+      sendLoopMessages();
```
bpas247: Is this to initially run it on startup?

features/loopMessages/index.js
```diff
@@ -0,0 +1,75 @@
...
+function isLoopMessage(messageToCheck, loopMessage, bot) {
+  return (
+    isMessageFromBot(bot, messageToCheck) &&
+    messageToCheck.content.trim() === getContent(loopMessage.content)
+  );
+}
+
+async function sendLoopMessage(client, loopMessage) {
+  const channel = client.channels.get(loopMessage.channelId);
+  const channelMessages = await channel.fetchMessages({ limit: 50 });
```
bpas247: I'm guessing if the loop message is past 50 messages, it probably doesn't matter if we delete it or not 😛 I'm not exactly sure how the reactiflux website is parsing the job posts, not sure if it would pick up the bots messages for the jobs channel specifically.

features/loopMessages/index.js
```diff
@@ -0,0 +1,75 @@
...
+ * 2. Fetch the last 50 messages of the channel
+ * 3. If the last message is already the message we want to loop, do nothing (we don't want to spam the channel)
+ * 4. Delete all loop messages (so the channel isn't filled with bot messages)
+ * 5. Send the loop message
+ */
+
+const loopMessages = require("./messages.js");
+
+// time between each check to send a new loop message
+const INTERVAL_TIME = 5 * 60 * 1000; // 5 minutes (ms)
```
bpas247: Is 5 minutes an appropriate amount of time? Should it be longer? I'm not sure if we want this to be customization depending on the loop message.

features/loopMessages/index.js
rafaelalmeidatk: Yep, if the bot restarts for some reason then it already executes the function to update the channels

features/loopMessages/index.js
rafaelalmeidatk: > I'm guessing if the loop message is past 50 messages, it probably doesn't matter if we delete it or not

Yeah exactly.

> I'm not exactly sure how the reactiflux website is parsing the job posts, not sure if it would pick up the bots messages for the jobs channel specifically.

I was gonna say that the bot only parses the messages that contain the tags `([FOR HIRE]`, `[HIRING]`, etc) but I just realized that they are in the message content lol
Great catch, I will change the jobs script to ignore the messages from the bot

features/loopMessages/index.js
rafaelalmeidatk: I am not sure, to be honest. The bot won't keep sending new messages if the last one is from it so a low interval time won't create spam, but the messages on #jobs aren't that active to hide the bot message if the interval time is too big. :thinking:

rafaelalmeidatk: > I'm not sure if we want this to be customization depending on the loop message.

The initial implementation had an interval field on each message, but I decided to remove it to avoid making it more complex than it is initially needed. We can add when we feel necessary.

vcarl:
I haven't reviewed this much yet, but my initial thoughts for the feature was just to have it be straight up a schedule of how often a message gets posted, and posting it. i was thinking [node-cron](https://www.npmjs.com/package/node-cron) for scheduling to keep it simple, not sure we need to check recent messages etc

rafaelalmeidatk:
@vcarl I see, so it would be definitely more simple hm :thinking: 
I like this implementation because the post is almost always visible to someone that is going to post something in the channel (most of them won't scroll up that much before posting something) and it doesn't spam the channel with the same message.

features/loopMessages/index.js
rafaelalmeidatk: I increased the interval to 40 minutes

rafaelalmeidatk: ok now each message can have a different interval

features/jobs.js
```diff
@@ -3,6 +3,8 @@ const cooldown = require("./cooldown").default;
 const jobs = {
   tags: ["forhire", "for hire", "hiring", "remote", "local"],
   handleMessage: ({ msg, user }) => {
+    if (msg.author.bot) return;
```
TrevorSayre: Is this strictly necessary?

bpas247: Yes? What benefit would we get if we had bots messaging/triggering each other?

rafaelalmeidatk: This is necessary because the bot may post a message in the `#jobs` channel that does not contain these tags so it would receive a warning from itself

rosofo:
May I ask why this wasn't merged? seems a waste

PR #25, opened 2019-11-25
feat(commands): add state updates may be async command

### LINKED ISSUE #24: Add state updates are async command
It should have a small blurb on state updates being async followed by a link to the [docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) (and additionally to this [gist](https://gist.github.com/bpas247/e177a772b293025e5324219d231cf32c) I wrote on it 😄)

bpas247:
Adds the "state updates may be async" command, for users who
get a *little* confused about the implications from it.

Fixes #24

### DIFF
--- a/features/commands.js
+++ b/features/commands.js
@@ -87,7 +87,7 @@ https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html`,
     },
     {
       words: [`!stateupdates`, `!su`],
-      help: `Explains the issue with state updates being asynchronous`,
+      help: `Explains the implications involved with state updates being asynchronous`,
       handleMessage: msg => {
         msg.channel.send({
           embed: {
--- a/features/commands.js
+++ b/features/commands.js
@@ -102,22 +102,8 @@ const handleEvent = e => {
 }
 \`\`\`
 where \`state\` is not the most up to date value when you log it. This is caused by state updates being asynchronous, so synchronous logic after a state update should not rely on the state value.
-\`\`\`js
-// log it directly inside render
-console.log(state);
 
-// Or in a useEffect call
-useEffect(() => {
-  console.log(state);
-}, [state]);
-
-const handleEvent = e => {
-  setState(e.target.value);
-};
-\`\`\`
 Check out these resources for more information:
-https://github.com/facebook/react/issues/11527#issuecomment-360199710
-https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973
 https://gist.github.com/bpas247/e177a772b293025e5324219d231cf32c
 https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous for more information.`,
             color: 7506394
--- a/features/commands.js
+++ b/features/commands.js
@@ -115,7 +115,11 @@ const handleEvent = e => {
   setState(e.target.value);
 };
 \`\`\`
-Check out https://gist.github.com/bpas247/e177a772b293025e5324219d231cf32c and https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous for more information.`,
+Check out these resources for more information:
+https://github.com/facebook/react/issues/11527#issuecomment-360199710
+https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973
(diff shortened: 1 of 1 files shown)


features/commands.js
```diff
@@ -85,6 +85,46 @@ https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html`,
...
+
+// Or in a useEffect call
+useEffect(() => {
+  console.log(state);
+}, [state]);
+
+const handleEvent = e => {
+  setState(e.target.value);
+};
+\`\`\`
```
vcarl: I feel like this misses the other solution of 
```
const handleEvent = e => {
  const {value} = e.target
  setState(value);
  console.log(value);
};
```
thoughts? It also feels a little long already, maybe drop the fixed code example and reduce how many links it includes? Defining the problem, then offering 3 solutions and 4 links to more resources seems like a lot.

features/commands.js
```diff
@@ -85,6 +85,46 @@ https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html`,
...
+    {
+      words: [`!stateUpdates`],
+      help: `Explains the issue with state updates being asynchronous`,
+      handleMessage: msg => {
+        msg.channel.send({
+          embed: {
+            title:
+              "State Updates May Be Asynchronous",
+            type: "rich",
+            description: `Often times you might run into an issue like this
```
bpas247:
```suggestion
            description: `Often times you run into an issue like this
```

features/commands.js
mokimo: Agree on this one:
>Defining the problem, then offering 3 solutions and 4 links to more resources seems like a lot.

bpas247: Okay, omitting the solution from the example makes sense, especially since my gist already covers all three solutions (and has a basic solution up front). Which links should I omit from it though? They all seem pretty valuable to include (I guess I *could* update the gist to link to Dan's resources under a new section, and then just include the original two links)

mokimo: Yeah using the gist and a `further reads on this topic (or sth..)` is probably the smartest idea

features/commands.js
```diff
@@ -85,6 +85,32 @@ https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html`,
         });
       }
     },
+    {
+      words: [`!stateupdates`, `!su`],
+      help: `Explains the issue with state updates being asynchronous`,
```
bpas247: there isn't really an issue with state updates being asynchronous, so I don't think `issue` is the right wording here.
```suggestion
      help: `Explains the implications involved with state updates being asynchronous`,
```
I'm not sure if this is any better? 🤔

features/commands.js
```diff
@@ -85,6 +85,32 @@ https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html`,
...
+              "State Updates May Be Asynchronous",
+            type: "rich",
+            description: `Often times you run into an issue like this
+\`\`\`js
+const handleEvent = e => {
+  setState(e.target.value);
+  console.log(state);
+}
+\`\`\`
+where \`state\` is not the most up to date value when you log it. This is caused by state updates being asynchronous.
```
bpas247: So the title of this is 
> State Updates May Be Asynchronous

But the wording here is 

> This is caused by state updates being asynchronous.

That's pretty confusing. The former statement is just a mirror of the [official docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) title and makes it sound uncertain whether or not that state updates are asynchronous, which seems to contradict the latter statement. 

I'm not sure which is more appropriate here 😬 I would assume the official docs are *probably* more credible, but I'm not sure how to re-word the latter statement (and I would also have to re-word my explanations in the gist that's linked here).

bpas247: Update: so I thought about it some more and realized that the reason why state in function components is always asynchronous is because of closures. A function component can only capture the state variable at any given time, and for the remainder of that closure it will always have that value that was captured.

This is different in the behavior of class components, as it's not dealing with closures. `this.state` *could* technically be updated synchronously, depending on when react ultimately decides to update it.

It's usually not good practice to expect state updates in class components to happen synchronously though, so best practice is to always treat state updates as asynchronous.

That being said, the wording here is probably fine since the gist is explicitly referring to state updates within function components. I might add a blurb with the difference in how state updates are handled within class components to the gist, as to provide more clarity on the subject.

PR #28, opened 2019-12-26
Add support for ⚠and 👎emoji moderation

dylangarcia:
⚠️ is for staff to log warnings for a user and their specific message
👎 is for the community to ping the moderators once a threshold (2 currently) is met

### DIFF
--- a/features/emojiMod.js
+++ b/features/emojiMod.js
@@ -83,10 +83,14 @@ const reactionHandlers = {
     }
   },
   "👎": (bot, reaction, message, member) => {
+    if (isStaff(member)) {
+      return;
+    }
+
     if (cooldown.hasCooldown(member.id, "thumbsdown")) {
       return;
     }
-    cooldown.addCooldown(member.id, "thumbsdown", 20 * 60);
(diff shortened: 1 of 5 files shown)


features/emojiMod.js
```diff
@@ -0,0 +1,165 @@
...
+    ]
+      .filter(Boolean)
+      .join("");
+
+    if (numberOfTotalReactions >= config.warningThreshold) {
+      logMessage = `<@${message.author.id}> has met the warning threshold in <#${message.channel.id}> for the message:`;
+    }
+
+    if (numberOfStaffReactions >= config.deletionThreshold) {
+      logMessage = `\`<@${message.author.id}> has met the deletion threshold in <#${message.channel.id}> for the message:`;
```
vcarl: Is this an unclosed ` at the very start of the message? Not sure if that's intentional.

dylangarcia: Looks like it, I originally had it as `@moderator` so it wouldn't ping during testing. Nice catch :eyes:

Issue #14, opened 2019-07-12
Add command to notify mods if a marked user reappears


vcarl:
One issue we've run into when moderating is problem users who leave on their own. It's great when they do, but there have been occasions where we want to keep an eye out if they come back. 

It would be great to have a command like `!watch` that would notify us if somebody resurfaces.  e.g. `!watch @user#1234 they were being super shady and might be scamming`. The next message @user#1234 sent would notify mods, then stop watching them.

I'm not sure how we'd persist this across bot reloads. Often we have people show back up after weeks/months away, so it's important it's persistent.

Padmanabh82:
@nickmccurdy do we need this ??? If not please close this

vcarl:
@Padmanabh82 I don't need help project-managing this repo, thanks.

Padmanabh82:
@nickmccurdy it's open from 3y, should we close it ?? As of now we don't need to watch anyone

PR #30, opened 2019-12-27
Automagically keep track of restricted users + add on rejoin

dylangarcia:
paaaawwwllllll

### DIFF
--- a/features/emojiMod.js
+++ b/features/emojiMod.js
@@ -1,3 +1,4 @@
+const knex = require("../db");
 const staffRoles = ["mvp", "moderator", "admin", "admins"];
 
 const isStaff = member =>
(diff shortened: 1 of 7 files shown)


features/emojiMod.js
```diff
@@ -129,36 +133,147 @@ const reactionHandlers = {
...
-        message,
-        message.guild.members.get(user.id)
+const emojiMod = bot => {
+  setTimeout(() => {
+    setupRestrictionsIntervals(bot);
+
+    setInterval(() => {
+      setupRestrictionsIntervals(bot);
+    }, 12 * 60 * 60 * 1000); // 12 hours
+  }, 10000);
```
thisRaptori: How come this is done in a `setTimeout`? 🤔

dylangarcia: @thisRaptori *sigh*

The `bot` isn't fully done setting up at that point, so it can't read members/servers immediately. It's _almost_ done setting up, but not quite, so there's the timeout. I tried like ~3-4 seconds but it wasn't 100% when joining RF.

If you want to (:eyes:) you have to hook it up to the lifecycle and initialize it in there, but afaict that would be a pretty decent restructure for this file that I deemed wasn't worth it.

thisRaptori: Yeah makes sense, no point completely redoing it all just for this 👍

nickmccurdy:
Is this a fix for #14?

nickmccurdy:
FWIW we can now use timeouts instead of re-restricting users, though it may still be useful for us to have notifications in case a lot of time has passed since the reported behavior.

thisRaptori:
image

PR #31, opened 2020-01-29
Shorten the description of !ask

nickmccurdy:
The example is too generic to be useful and doesn't have an actionable question, meanwhile the code help isn't as detailed as `!code` and is just taking up extra space

### DIFF
--- a/features/commands.js
+++ b/features/commands.js
@@ -178,19 +178,7 @@ https://reactjs.org/docs/lifting-state-up.html`,
           embed: {
             title: "Asking to ask",
             type: "rich",
-            description: `Instead of asking to ask, ask your question instead. People can help you better if they know your question.
-
-Example:
-
-Bad: "Hey can anyone help me with some JS?"
-Bad: "Anyone good with JS?"
-Good: "I'm having trouble adding a class to a div using JS. Can I have some help?"
-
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Any suggestions to change the first sentence? Personally I don't think the removed text provides any more friendless/reassurance though.

vcarl:
I feel like it's a little short to merit a bot response—this wouldn't be hard to type out manually and skip the whole impersonal aspect. The sentence + some examples I think would strike a balance, and can strip some lines to make the overall thing shorter. Maybe:

```
Instead of asking to ask, ask your question instead.

Bad: "hey can anyone help me?"	
Bad: "anyone good with redux?"
Good: 
> I'm trying to fire a redux action from my component, but it's not getting to the reducer.
> \`\`\`js
> // snippet of code
> \`\`\`
> I'm seeing an error, but I don't know if it's related.
> \`Uncaught TypeError: undefined is not a function\`
```

nickmccurdy:
I added the example, but I still think this line is important as it explains why this is suggested:
> People can help you better if they know your question.

nickmccurdy:
I just discovered this site, maybe we could link to it as an example instead? https://www.dontasktoask.com/

vcarl:
Sounds good to me!

PR #32, opened 2020-01-31
Add !crosspost/!crossposting command

nickserv:
Related to #15 

This is an interim solution to something more automated like #17.


features/commands.js
```diff
@@ -306,6 +306,20 @@ Here's an article explaining the difference between the two: https://goshakkk.na
 
         fetchMsg.delete();
       }
+    },
+    {
+      words: [`!crosspost`, `!crossposting`],
```
bpas247: maybe we could add a shorter alias? `!cross` perhaps?

vcarl: or
```suggestion
      words: [`!crosspost`, `!xpost`],
```

features/commands.js
```diff
@@ -306,6 +306,20 @@ Here's an article explaining the difference between the two: https://goshakkk.na
...
+      help: `explains why and how to avoid crossposting`,
+      handleMessage: msg => {
+        msg.channel.send({
+          embed: {
+            title: "Crossposting",
+            type: "rich",
+            description: `Do not crosspost messages by copying and pasting them into multiple channels. This makes it harder to help you as discussions may be broken up into multiple channels, and takes more space from other users asking questions. Instead, post your message once to the channel you think is most relevant.`,
```
vcarl: Maybe soften it a bit, be a little less specific.
```suggestion
            description: `Please don't ask the same question in multiple channels. It makes it harder to help you by splitting the discussion into multiple places, and takes more space from other users asking questions. Instead, post your message once to the channel you think is most relevant.`,
```

nickmccurdy: I prefer having a more generalized definition of crossposting involving posting messages in multiple channels. Asking a question in multiple channels is one form of crossposting, but I'd also like to use this for things like spamming, requests for feedback, notes, and duplicate links.

nickmccurdy:
Merged and ported to TypeScript

nickmccurdy:
Looking for reviews

features/commands.js
nickmccurdy: @vcarl Thoughts on this? We can still soften the language but I think it's important to not only recommend against crossposting questions.

Padmanabh82:
@nickmccurdy why don't you merged this ???

PR #34, opened 2020-02-12
Add !appIdeas to command bot.

mlynchdev:
Myself and others in the community have reviewed this github link and find it very helpful.  I think it could benefit the ecosystem and honestly at least a couple times a day a new Reactiflux-er is asking "What can I build?".  This is a helpful list, non opinionated and well documented with user stories etc.  I think it would be a benefit to the Reactiflux community.

-Mark "Elsewhere" Lynch

### DIFF
--- a/features/commands.js
+++ b/features/commands.js
@@ -307,7 +307,7 @@ Here's an article explaining the difference between the two: https://goshakkk.na
       }
     },
     {
-      words: [`!appIdeas`],
+      words: [`!appideas`],
       help: `provides a link to the best curated app ideas for beginners to advanced devs`,
       handleMessage: msg => {
         msg.channel.send({
--- a/features/commands.js
+++ b/features/commands.js
@@ -306,7 +306,26 @@ Here's an article explaining the difference between the two: https://goshakkk.na
 
         fetchMsg.delete();
       }
-    }
+		},
+		{
+      words: [
+        `!appIdeas`,
+      ],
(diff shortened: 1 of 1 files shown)


features/commands.js
```diff
@@ -306,7 +306,26 @@ Here's an article explaining the difference between the two: https://goshakkk.na
 
         fetchMsg.delete();
       }
-    }
+		},
+		{
```
nickmccurdy: Please fix the indentation on these couple lines to make them consistent

mlynchdev: Yeah, I'm okay with this.

nickmccurdy:
I've updated our Prettier formatting and added the default config in master. Please merge it into your branch, rerun Prettier, and commit to the pull request.

features/commands.js
```diff
@@ -305,6 +305,23 @@ Here's an article explaining the difference between the two: https://goshakkk.na
 
         fetchMsg.delete();
       }
+    },
+    {
+      words: [`!appIdeas`],
```
nickmccurdy:
```suggestion
      words: [`!appideas`],
```I looked over our current code again, and realized that the capital `I` here could break our case sensitivity checks, because the trigger isn't lowercased while comparing against the lowercased message. Making this change would allow both `!appideas` and `!appIdeas`

mlynchdev:
Thanks for your patience!  Glad to help!

PR #37, opened 2020-02-14
Add package lock

nickmccurdy:
This is a similar idea to #35 but using the same version of discord.js that the bot is currently using.

I copied the package-lock.json directly from the server to prevent interruptions, though it's using a slightly older version of npm (6.11.3). If the lockfile becomes too inconsistent we may need to update the server's npm version.


nickmccurdy:
Got approval from @dylangarcia on Discord, needs one more review

PR #57, opened 2020-05-03
Add !immutability command

dylangarcia:
image

yer welcome @markerikson

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -371,8 +371,8 @@ Here's an article explaining the difference between the two: https://goshakkk.na
         embed: {
           title: "Immutability",
           type: "rich",
-          description: `Immutability is the concept of updating data by creating entirely new variables that do not have references to previous data.
-          You can not modify existing data directly and must update the entire object.  
+          description: `Immutability is the concept of updating data by creating new values rather than mutating values at existing references.
+          You cannot modify existing data directly in React or Redux.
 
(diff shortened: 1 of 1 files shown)


src/features/commands.ts
```diff
@@ -362,6 +362,25 @@ Here's an article explaining the difference between the two: https://goshakkk.na
...
+  {
+    words: [`!immutability`],
+    help: `provides resources for helping with immutability`,
+    handleMessage: msg => {
+      msg.channel.send({
+        embed: {
+          title: "Immutability",
+          type: "rich",
+          description: `Immutability is the concept of updating data by creating entirely new variables that do not have references to previous data.
+          You can not modify existing data directly and must update the entire object.
```
nickmccurdy, line 375: # Notes
- Immutability does not require creating new variables, just new variables
- Immutable updates often include existing references
- `can not` => `cannot`
- The entire object doesn't need to be updated if you can get away with a shallow update
- Clarifying that immutability is often required with React and Redux but not necessarily other JS
```suggestion
          description: `Immutability is the concept of updating data by creating new values rather than mutating values at existing references.
          You cannot modify existing data directly in React or Redux.
```

nickmccurdy:
@markerikson Do you think this could potentially confuse Redux Toolkit users, and if so do you have any suggestions?
> You cannot modify existing data directly in React or Redux.

PR #58, opened 2020-05-04
Add !formatting command

nickmccurdy:
image


nickmccurdy:
That's what I was trying to suggest with "create churn from style debates" vs."follow best practices. I wouldn't want to get too wordy with in-depth explanations, but maybe we could link to a docs page regarding this? https://prettier.io/docs/en/index.html

vcarl:
In the interest of getting this out I'm gonna merge. Addressing anti-formatter arguments is a nice-to-have to be sure.

PR #60, opened 2020-05-04
Use MDN search API for !mdn

vcarl:


src/features/commands.ts
```diff
@@ -396,7 +390,7 @@ const commands: ChannelHandlers = {
       });
 
       if (keyword) {
-        if (cooldown.hasCooldown(msg.author.id, `commands.${keyword}`)) return;
+        // if (cooldown.hasCooldown(msg.author.id, `commands.${keyword}`)) return;
```
nickmccurdy, line 393: Better to delete than comment, we have version control ```suggestion
```

Comment (vcarl) [reply src/features/commands.ts:393]: Oops this was a dev productivity change, didn't mean to commit

Comment (vcarl) [inline src/features/commands.ts:393]:
```suggestion
        if (cooldown.hasCooldown(msg.author.id, `commands.${keyword}`)) return;
```

PR #64, opened 2020-06-24
Set up CI with GitHub Actions

nickmccurdy:


nickmccurdy:
Not sure if we want to build or run the Docker image in CI, feedback welcome

bpas247:
> Not sure if we want to build or run the Docker image in CI

I don't think there's much benefit to that, since we're already running those same commands in this workflow

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - run: npm ci
+    - run: npm test
+    - run: npm run build --if-present
```
bpas247, line 14: Do we want names for these steps? might make them more readable in the logs

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,14 @@
+name: Node.js CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
```
bpas247, line 6: We should add a name to this job

nickmccurdy: I thought this key was the job name. If you check the action run, it's listed as `build` on the sidebar, but you can change it to something else and that will appear.

bpas247: yes but we can still supply it with a more descriptive name by adding a [name](https://github.com/react-bootstrap/react-bootstrap/blob/f14ca04848bbfe52b6655e9bba5408ec87a53c8e/.github/workflows/ci.yml#L13) property to it (granted that link isn't a great example 😆)

nickmccurdy: What would it be other than CI? That makes sense to me

bpas247: lol true. I'll go ahead and approve the PR either way since I don't want this to a blocker

nickmccurdy:
I want to try to merge this with the existing workflow before merging, so we can enforce deploys only after CI passes

nickmccurdy:
I think it's working! Can I get a quick sanity check?

.github/workflows/node.js.yml
```diff
@@ -1,15 +1,26 @@
...
+      - name: Install
+        run: npm ci
+      - name: Test
+        run: npm test
+      - name: Build
+        run: npm run build --if-present
+
+  deployment:
+    needs: build
+    if: github.ref == 'master' || github.ref == 'develop' || github.ref == 'feature/actions'
```
bpas247, line 21: Do we have a `feature/actions` branch? or is this something setup for strictly testing our CI?

dylangarcia: It was for testing the branch that implemented the first action without having to push directly to master/dev because we're not degenerates

nickmccurdy: That was from the old deployment workflow. I'm assuming the purpose is to be able to deploy updates to the workflow without merging into master. It also let me test this PR out, and I finally got the deploy working!

nickmccurdy:
Can confirm this deploys successfully (the workflow runs CI and deployment, and then Reactibot logs to us that it's started on Discord)

nickmccurdy:
Anyone else want to review before we merge?

PR #70, opened 2020-10-25
Warn users when they try to ping here or everyone

dylangarcia:
image

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -431,7 +431,7 @@ To integrate it into your editor: https://prettier.io/docs/en/editors.html`,
         embed: {
           title: "Tsk tsk.",
           type: "rich",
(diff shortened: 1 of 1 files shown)


src/features/commands.ts
```diff
@@ -410,6 +412,30 @@ To integrate it into your editor: https://prettier.io/docs/en/editors.html`,
...
+
+      if (!member || isStaff(member)) {
+        return;
+      }
+
+      msg.channel.send({
+        embed: {
+          title: "Tsk tsk.",
+          type: "rich",
+          description: `Please do **not** try to use \`@here\` or \`@everyone\` - there are ${msg.guild.memberCount} members in Reactiflux. Everybody here is a volunteer, and somebody will help you if they can.`,
```
bpas247, line 434: Although some people also try to ping everyone when they've made something and they're trying to share it in `#i-made-this`, maybe we could also add a short blurb for that scenario?

dylangarcia: `Somebody will respond if they can`?

bpas247: works for me

bpas247 (approved):
🚢

PR #74, opened 2020-11-08
Add snack.io to code options

rahsheen:


nickmccurdy:
Good idea!

Is it supposed to default to web or should be override that?

rahsheen:
I think it always defaults to web, but maybe there's a way around that. Will check.

nickmccurdy:
You're more familiar with React Native than I am. I'm not just sure if that's common for React Native Expo developers, or if it's specifically for web developers.

nickmccurdy:
We can merge this and figure out the default mode thing later if it's an issue.

Issue #68, opened 2020-09-18
Warn reaction fails on messages > Discord message limit


vcarl:
We should truncate large messages that would put the bot log message over the character limit

vcarl:
Fixed in #79

PR #79, opened 2020-12-04
DM members who receive a ⚠️ warning from moderators

### LINKED ISSUE #78: DM members who receive a ⚠️ warning from moderators
The ⚠️ reaction can only be triggered by moderators. Currently it updates #mod-log with a copy of the message, but it should also DM the member who sent the message. I'm imagining something like

> You've received a warning from the moderators on your message in #channel: 
> ```
> message content
> ```
> link: https://discord.com/channels/xxx/yyy/zzz

In the future, maybe we could expand this to have a "discussion channel" for the user to privately discuss the situation with all mods.

Huuums:
- fix #78 
- Added a very handholdy contributing.md to set up the bot

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+# Contributing
+
+## Setting up the bot
+
+1. Fork the repository and clone it to your workstation.
+2. Create a `.env` file and copy the contents of `.env.example` into it.
+3. Go to your [Github Settings](https://github.com/settings/tokens) and create a personal access token with the scope "gist"
+4. Copy the token and put it into the `.env` file as `GITHUB_TOKEN`
+5. [Create a new Discord application](https://discord.com/developers/applications)
+6. Switch to the "Bot" tab and click "Add Bot"
+7. Toggle off the "Public Bot" switch and save.
+8. Copy the token and put it into the `.env` file as `DISCORD_HASH`.
+9. Go to the OAuth2 Tab and choose the "bot" scope.
+10. Check the "Administrator" permission.
+11. Copy the url out of the "Scopes" Panel and paste it into your browser.
+12. (We recommend to create a new Discord Server before Step 12)
+13. Select the server to add your bot.
+14. run `npm install` and then `npm run dev`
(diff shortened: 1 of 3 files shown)


Huuums:
after checking out the other issues I also added a fix for #68

src/features/emojiMod.ts
```diff
@@ -97,6 +98,17 @@ const reactionHandlers: ReactionHandlers = {
...
     }
+
+    const privateMessageToSender = [
+      `You've received a warning from the moderators on your message in ${message.channel}`,
+      "\n\n",
+      "Your message: \n",
+      `\`${truncateMessage(message.content)}\``,
+      "\n\n",
+      `Link: https://discord.com/channels/${message.guild?.id}/${message.channel.id}/${message.id}`
+    ].join("");
```
vcarl, line 109: Just curious, why not a multiline string? This seems much noisier
```suggestion
    const privateMessageToSender = 
      `You've received a warning from the moderators on your message in ${message.channel}

Your message:
\`${truncateMessage(message.content)}\`

Link: https://discord.com/channels/${message.guild?.id}/${message.channel.id}/${message.id}
```

Huuums: There's no specific reason. To be honest I just didn't think about it.

https://github.com/reactiflux/reactibot/blob/546418eea4b7d99a2cc4b324cd141a703a13142e/src/features/emojiMod.ts#L67

I copied this message and modified the text a bit ignoring the fact that there's some filtering going on in the example.

CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
+# Contributing
```
vcarl, line 1: :100: this is super awesome to have, thanks for adding it.

src/utils.ts
```diff
@@ -9,3 +9,15 @@ export const isStaff = (member: GuildMember | null | undefined) => {
     staffRoles.includes(role.name.toLowerCase())
   );
 };
+
+//Discords limit for messagelength
```
vcarl, line 13: just gonna
```suggestion
// Discord's limit for message length
```
real quick

vcarl (approved):
I think I'd like to adjust the format of the message we send (a ` within the message I think will bork the formatting as-is) but that's not a blocker for merging. Thanks for doing this!!

PR #85, opened 2021-01-07
log bans to modlog channel

### LINKED ISSUE #45: Have bot ping bans/kicks to #mod-log

Huuums:
this is a PR to fix #45 partially. 

Unfortunately it's not possible to get the info of who banned the user. 

It doesn't seem to be possible to log kicks specifically. There is a [guildMemberRemove ](https://discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-guildMemberRemove) event but that seems to trigger everytime a user leaves or gets kicked. You can't distinguish the two. 

I tried to use the `deleted` property on the GuildMember object that gets passed in the guildMemberRemove event but it seems to be false every time.

### DIFF
--- a/src/features/emojiMod.ts
+++ b/src/features/emojiMod.ts
@@ -1,6 +1,6 @@
-import { MessageReaction, Message, GuildMember, TextChannel } from "discord.js";
+import { MessageReaction, Message, GuildMember } from "discord.js";
 import cooldown from "./cooldown";
-import { isStaff, truncateMessage } from "../utils";
+import { getModLogChannel, isStaff, truncateMessage } from "../utils";
 import { ChannelHandlers } from "../types";
 
 const config = {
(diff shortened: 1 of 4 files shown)


taranvohra:
One possible way is, you can queue an entry on `guildBanAdd` event and then on `guildMemberRemove` you scan through the queue for a possible match and if it's found you can log it as a `BAN` in the modlog channel.

Huuums:
So after Iggy pointed out that you can in fact figure out who banned the person via getting the Auditlogs I improved the ban messages. 

I also used the Auditlog functionality to figure out if a user was kicked whenever a user leaves the server.  While it works as intended I feel that this is rather hacky and the Bot needs this permission so that it can work. [see yellow arrow](https://github.com/discordjs/discord.js/issues/5012#issuecomment-727269731).

I'm open to more feedback :)

vcarl:
Closing cuz this just never got merged and is now out of date :pensive: I believe there's (still) an issue with ban tracking where the audit log API does not reliably fire events, when I've tested in the past a reasonably high % of bans wouldn't get caught. I saw similar when testing a deleted message tracker but worse, something like 40-60% of deleted messages wouldn't fire a gateway event. This was also opened at a time when I personally was paying ~ the least attention I ever have to Reactiflux — sorry for letting this wither on the vine

PR #88, opened 2021-02-03
Message Scheduling

rosofo:
This PR introduces the ability to schedule messages using cron syntax.

To schedule a message you add an entry to `MESSAGE_SCHEDULE` with the following properties:

- The cron string
- An array of guilds. Inside each guild is its id and the ids of the channels to send to
- The message


rosofo:
https://github.com/reactiflux/reactibot/issues/13#issue-467624632

vcarl (approved):

PR #89, opened 2021-02-17
Switch !mdn to use mdn_url response field instead of slugs

mszabo-wikia:
Looks like MDN's search API behavior changed, so using the slug field to build an MDN article path is no longer sufficient (slugs are missing /docs prefix). As a fix, use the mdn_url field from the search API response instead, which seems more like what we need.


nickmccurdy (approved):

nickmccurdy:
This seems to be working smoothly now, thank you!

Issue #95, opened 2021-06-29
Prevent channels being used for bot testing


louis-young:
At the moment, people commonly mistake the `#random` channel for a bot testing channel (or simply choose to test bot commands there).

I think we could solve this by sending a friendly reminder to the user when bot commands are used in this channel.

PR #96, opened 2021-06-29
Feature/#95: Prevent channels being misused for bot testing

louis-young:
PR for #95 

This pull request helps to prevent channels from being misused for bot testing.

### Changes

- Add a list of channels in which the bot is disabled.
- Send a direct message to the user prompting them to test bot commands in direct messages.
  - If the user has direct messages from other users disabled on the server, notify them via the channel they triggered the bot command in.
  
### Notes
  
 - This is my first time ever looking at a Discord bot so please review with extra caution.
 - I had to cast the channel object to a `TextChannel` to access the `name` property. I don't like this, but this appears to be standard from what I could find on some Discord.js issues.

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -2,7 +2,11 @@ import fetch from "node-fetch";
 import { Message, TextChannel } from "discord.js";
 import cooldown from "./cooldown";
 import { ChannelHandlers } from "../types";
-import { isStaff } from "../utils";
+import {
+  isStaff,
+  isBotDisabledInChannel,
+  notifyUserBotResponseWasPrevented
+} from "../utils";
 
 const EMBED_COLOR = 7506394;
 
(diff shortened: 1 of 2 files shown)


nickmccurdy:
I understand the issue and I appreciate the effort, but I think that entirely disabling all bot commands in some channels restricts the usefulness of the bot too much. For example the `!gender` command is meant to be used as a response to a user misgendering, which can happen in any channel, including random. If you'd like to try, it would be nice to also have a list of commands that specifically get restricted.

louis-young:
> I understand the issue and I appreciate the effort, but I think that entirely disabling all bot commands in some channels restricts the usefulness of the bot too much. For example the `!gender` command is meant to be used as a response to a user misgendering, which can happen in any channel, including random. If you'd like to try, it would be nice to also have a list of commands that specifically get restricted.

That should be pretty trivial to do, I’ll take a stab in the morning 🙂

awareness481:
Is accepting the communication commands enough?

Something like this should work imo:

```js
const whitelistedCommands = ["!gender, !xy"];

export const isWhitelistedCommand = (commands: string[]) => {
  return (
    commands.filter(command => whitelistedCommands.includes(command)).length > 0
  );
};
//...
const hasPreventedBotResponse = isBotDisabledInChannel(channelName);
const isNotWhitelisted = !isWhitelistedCommand(command.words)

if (hasPreventedBotResponse && isNotWhitelisted) {
  notifyUserBotResponseWasPrevented(msg, channelName);

  return;
}
```

vcarl:
Maybe a different approach would be to change how the cooldown works? Right now it's hard coded with a specific duration, we could make it an exponential backoff situation so folks get blocked for longer times if they keep spamming them, with a DM'd message like "it definitely works, you can DM here with no cooldown if you want"

vcarl:
I'm going to close this out, since it's pretty stale (sorry!) and I think a different approach would be more successful. Let me know if you want to talk it through more, I like the spirit of this but don't think this is mergable as-is

PR #98, opened 2021-07-16
WIP   warning when steamcommunity word appears

Unbuttun:
1. The bot reacts with ⚠ when word steamcommunity appears, then do the mod-log thing.
2. Updated dotenv to v10 because it has types and no breaking changes for node 12.
3. Little prettier config to keep everything consistent as it had no config beforehand.


src/features/emojiMod.ts
```diff
@@ -163,24 +172,36 @@ const emojiMod: ChannelHandlers = {
...
-      !message.guild ||
-      user.id === bot.user?.id ||
-      message.author.id === bot.user?.id
-    )
+      !message.guild
+      // ||
+      // user.id === bot.user?.id ||
+      // message.author.id === bot.user?.id
+    ) {
+      console.log("this");
```
BTMPL, line 180: This ;)

src/features/emojiMod.ts
```diff
@@ -24,19 +30,22 @@ type ReactionHandlers = {
...
     // Skip if the user that reacted isn't in the staff or the post is from someone
-    // from the staff
+    // from the staff but not when it's done by the bot itself
+
     if (
-      !message.guild ||
-      !message.author ||
-      !isStaff(member) ||
-      isStaff(message.guild.member(message.author.id))
+      isBot(member, bot) &&
```
BTMPL, line 44: Are we sure we want `&&` here?

src/features/emojiMod.ts
```diff
@@ -163,24 +172,36 @@ const emojiMod: ChannelHandlers = {
...
 
     let emoji = reaction.emoji.toString();
+    const member = message.guild.member(user.id);
+    if (!member) return;
+
+    if (emoji === "⚠") {
+      const reactionHandler = reactionHandlers["⚠️"];
+
+      if (reactionHandler) {
+        reactionHandler(reaction, message, member, bot);
```
BTMPL, line 183: Will this not trigger the reactionHandler twice? Here (192) and 203?

nickmccurdy:
> Little prettier config to keep everything consistent as it had no config beforehand.

I appreciate the effort on improving this, but the config this PR added was equivalent to the defaults, and our empty config was already using the defaults. I've reverted this to keep the PR more focused, but if you want to change the defaults to something else feel free to submit another PR.

package.json
```diff
@@ -18,7 +18,7 @@
   "dependencies": {
     "@types/node-cron": "^2.0.3",
     "discord.js": "^12.5.1",
-    "dotenv": "^6.1.0",
+    "dotenv": "^10.0.0",
```
nickmccurdy, line 21: Note: This may seem like it's not doing much, but v6 doesn't include types so it actually errors the build in CI

src/features/emojiMod.ts
```diff
@@ -49,7 +58,7 @@ const reactionHandlers: ReactionHandlers = {
 
     const modLogChannel = message.guild?.channels.cache.find(
       channel =>
-        channel.name === "mod-log" || channel.id === "257930126145224704"
+        channel.name === "mod-log" || channel.id === "865641297599004713"
```
nickmccurdy, line 61: What are the old and new IDs referring to here? Neither of these match the ID of `mod-log` on Reactiflux, and I can't seem to find them via our API tokens either.

PR #100, opened 2021-07-17
Warn messages that ping everyone

dylangarcia:
I don't have my tokens on this computer but this should work.


nickmccurdy:
I'm testing this out, and while the bot does warn, it does not trigger the warn emote reaction. I think the issue is that the bot account itself is excempt from all emoji handlers: https://github.com/reactiflux/reactibot/blob/5aaa6d76eae30bc26bb1b855a5ee20e442fcb8f1/src/features/emojiMod.ts#L167

Do you think it would be safe for us to remove this line from the conditional check, or do we need to let this exception to the rule through a different way?

bpas247 (approved):

PR #114, opened 2021-10-30
Prevent non-staff from reacting with ⚠️ and send a clarifying DM

Retsam:
We silently ignore ⚠️ reactions if they aren't from staff, which can make it hard for the user to tell if they've done anything or not.  This changes the logic to remove the ⚠️ reaction and DM the user explaining why it was removed.

(We talked about counting ⚠️ reactions from staff as if they were 👎, but is simpler to implement and I think ultimately clearer in result)


dylangarcia (approved):

vcarl:
💯 TY!

PR #116, opened 2021-10-31
Add !learn command and add beta docs to !hooks command

rahsheen:


src/features/commands.ts
```diff
@@ -566,7 +566,10 @@ To integrate it into your editor: https://prettier.io/docs/en/editors.html`,
           The official React docs are the best resource for learning hooks:
           https://reactjs.org/docs/hooks-intro.html
 
-          However, the React docs still teach classes in the tutorials. A rewrite is in progress, but until then, there's a "React with Hooks" version of the React docs that uses hooks and function components for all examples:
+          However, the React docs still teach classes in the tutorials. A rewrite is in progress here: 
+          https://beta.reactjs.org
+
+          There's also a "React with Hooks" version of the React docs that uses hooks and function components for all examples:
```
ByronGuina, line 572: Can probably roll this comment into the above sentence.

> However, the React docs still teach classes in the tutorials. A rewrite focused on functional components and hooks is in progress here:

src/features/commands.ts
```diff
@@ -577,6 +580,30 @@ To integrate it into your editor: https://prettier.io/docs/en/editors.html`,
...
+          title: "Learning React",
+          type: "rich",
+          description: `
+          The official (beta) React docs are the best resource for learning React:
+          https://beta.reactjs.org
+
+          The official React docs still teach classes for the examples, but the concepts are still valid:
+          https://reactjs.org/docs/getting-started.html
+
+          There's also a "React with Hooks" version of the React docs that uses hooks and function components for all examples:
```
ByronGuina, line 599: Same thing as the above comment.

> The official (beta) React docs are the best source for learning React. They focus on functional components and hooks:

rahsheen: It's something I wasn't sure about. My original thought was to just alias `!hooks` as `!learn`, but wasn't sure if others agreed there's kinda no reason to explicitly highlight hooks anymore.

ByronGuina: Yeah I think that makes sense to me. I think hooks have been the de facto mechanism for long enough, especially when the command is like 99% the same content just arranged differently.

BTMPL (approved):

PR #120, opened 2021-11-29
"tsk tsk" warning message thresh hold added

Padmanabh82:
Hi, staff members

I am Padmanabh. I have tried to made solution on spam problem

this solution will not disturb server for spamming @ here or @ everyone


Padmanabh82:
I hope @nickmccurdy will merge this pr

Padmanabh82:
And just check code from your side, from my side it is good

.env.example
```diff
@@ -1,2 +0,0 @@
-DISCORD_HASH=
```
ghardin137, line 1: why did you remove these?

package.json
```diff
@@ -21,7 +21,7 @@
     "dotenv": "^6.1.0",
     "gists": "^2.0.0",
     "node-cron": "^2.0.3",
-    "node-fetch": "^2.2.0",
+    "node-fetch": "^2.6.6",
```
ghardin137, line 24: This change was unnecessary. the old version matches this version as well

src/features/commands.ts
```diff
@@ -609,15 +609,36 @@ Instead:
...
+      await msg.channel.messages.fetch({ limit: 10 }).then((messages) => {
+        let SendInPublic = true
+        messages.map((key) => {
+          if (key.content.startsWith("Please do **not** try to use \`@here\` or \`@everyone\` - there are ")) {
+          SendInPublic = false
+          }
+        });
+
+        if(SendInPublic != true){
+          msg.author.send({
```
ghardin137, line 621: This won't work if the user has DMs disabled or has left the server.

src/features/commands.ts
```diff
@@ -609,15 +609,36 @@ Instead:
...
       await msg.react("⚠️");
 
-      await msg.reply({
-        embed: {
-          title: "Tsk tsk.",
-          type: "rich",
-          description: `Please do **not** try to use \`@here\` or \`@everyone\` - there are ${msg.guild.memberCount} members in Reactiflux. Everybody here is a volunteer, and somebody will respond when they can.`,
-          color: "#BA0C2F",
-        },
+      await msg.channel.messages.fetch({ limit: 10 }).then((messages) => {
```
ghardin137, line 612: a better option here would be to keep track of the timestamp of the last time it replied with this. and only send the message if that time difference is greater than a given threshold

ghardin137 (changes requested):

package.json
Padmanabh82: I did edit this

.env.example
Padmanabh82: I didn't touch this

src/features/commands.ts
Padmanabh82: Ok sir I will use that system

src/features/commands.ts
Padmanabh82: So, i think we can only delete it, no sending message

src/features/commands.ts
Padmanabh82: @ghardin137

Padmanabh82: @ghardin137

Padmanabh82: @ghardin137 but what should be timeout ??? and it should be in  milliseconds, seconds or minutes?????

thisRaptori (commented):
I haven't run this, but based on reading the code a) I'm pretty sure it doesn't do what you intended, and b) it doesn't actually implement the thing Hums was talking about.

For a), the if statement condition makes no sense. `Last_message_time.getMinutes() <= threshold`: - you're checking whether the number of minutes since the most recent hour (i.e. 4:32 -> 32, 6:21 -> 21, etc) is less than or equal to 0... So this code path will run only during the first minute of each hour.

For b), the suggestion in the screenshot was for the user to be timed out, not for the bot's warning message to be timed out, so this isn't actually what was suggested!

Also, you've introduced a bunch of unnecessary changes which should not be part of this PR! Would question whether we do actually want to prevent the bot from posting its warning in each relevant channel, since it can be useful to have the message there to remind users of the rule.

Padmanabh82:
@thisRaptori oooh sir, my english problem. So, what should I do now ??? Should I consider your point a and develop it ???

thisRaptori:
I think we'd need to consider whether it's something we actually want in the bot. Carl might be putting some time into figuring that out and rebuilding/refactoring some of the bot, so best to wait until after that's done before looking at any open feature requests which are left out of the initial refactor.

Closing this for now because of the above!

PR #122, opened 2021-12-04
Debounce spamlogs

vcarl:
I refactored the emoji mod feature a good bit, cuz it ended up being quite annoying to add a debounce with the previous structure. Couldn't rely on the message being stable, so debounce wouldn't have trigged correctly. Now message creation is separate, as is post new/edit existing mod log message

Before:
Screen Shot 2021-12-07 at 4 26 36 PM

After:
Screen Shot 2021-12-07 at 4 26 43 PM


src/features/emojiMod.ts
```diff
@@ -28,6 +31,27 @@ type ReactionHandlers = {
...
+);
+const handleReport = (
+  channelInstance: TextChannel,
+  reportedMessage: Message,
+  logBody: string,
+) => {
+  if (warningMessages[reportedMessage.id]) {
+    warningMessages[reportedMessage.id].edit(logBody);
+  } else {
+    sendModLog(channelInstance, logBody).then((warningMessage) => {
```
vcarl, line 49: oops, I need to make sure this updates the cached list. forgot to copy off after splitting it out from sending

vcarl:
The channel listed in the warning (and the link to the message) won't be complete, but I think that's probably fine. Our biggest problems with spammers are already auto deleted by the bot, so we don't actually need to know about them

PR #124, opened 2021-12-07
Add recurring messages to some channels

vcarl:
Closes #124. We've had support for this forever, but have never actually used them! This re-posts rules to `#job-board` every day, and some other messages intended to guide new joiners in our highest traffic channels.


src/features/scheduled-messages.ts
```diff
@@ -1,15 +1,35 @@
...
 
+const enum CHANNELS {
+  "helpReact" = "103696749012467712",
+  "helpJs" = "565213527673929729",
+  "random" = "103325358643752960",
+  "jobBoard" = "103882387330457600",
+}
+
+// By keeping these at odd divisions, we can make sure they show up at all timezones. If it were */24, for instance, it would consistently show up in the middle of the night for some timezones.
+const enum FREQUENCY {
```
vcarl, line 13: Instead of cron, we should swap this out to be a zeroed-out timer instead. On bot startup, init relative to Sunday midnight and configure intervals against that. cron doesn't have a way of doing cross-day time durations, like sending a message every 18 hours.

ghardin137 (approved):

src/features/scheduled-messages.ts
```diff
@@ -21,13 +39,77 @@ export const MESSAGE_SCHEDULE: MessageConfig[] = [
     }
   }
   */
-];
+  {
+    interval: FREQUENCY.daily,
+    postTo: [{ channelId: CHANNELS.jobBoard }],
```
dylangarcia, line 44: Would it be worth combining these two properties so you could do something like this? We could have the same message on different intervals for different channels.

```js
postTo: [
   { channelId, interval },
]```

Comment (vcarl) [reply src/features/scheduled-messages.ts:44]: Oh I'm into that

File: src/features/scheduled-messages.ts [hunk 3]
```diff
@@ -36,43 +118,39 @@ export const scheduleMessages = (
   messageConfigs: MessageConfig[],
 ) => {
   const scheduledTasks = messageConfigs.map((messageConfig) =>
-    scheduleMessage(bot, messageConfig),
+    scheduleTask(messageConfig.interval, () => {
```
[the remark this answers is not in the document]
vcarl: I briefly tried, but the most obvious way produced a promise that wouldn't resolve for multiple hours which seemed bad. And since it's a recurring task, a single promise resolution doesn't really make sense

src/features/scheduled-messages.ts
vcarl: Pushed

src/features/scheduled-messages.ts
```diff
@@ -1,85 +1,152 @@
...
-): channel is discord.TextChannel | discord.DMChannel | discord.NewsChannel => {
-  return (
-    channel.type === "GUILD_TEXT" ||
-    channel.type === "GUILD_NEWS" ||
-    channel.type === "GUILD_STORE"
-  );
+export const scheduleMessages = (bot: discord.Client) => {
+  bot.on("ready", () => {
+    MESSAGE_SCHEDULE.forEach((messageConfig) =>
+      sendMessage(bot, messageConfig),
```
vcarl, line 119: Turns out the bug was that I was trying to send messages before the bot logged in, not anything related to permissions/etc 🤦 oh well, we got there

src/features/scheduled-messages.ts
```diff
@@ -1,85 +1,152 @@
...
+* Posts should be reasonably descriptive.
+* Jobs are paid — unpaid, equity-only, or similar are not allowed.
+* We don't allow "small gigs".
+
+Moderators may remove posts at any time, with or without warning. Repeat violators of these rules will be removed from the server permanently.
+`,
+    },
+  },
+  {
+    postTo: [{ interval: FREQUENCY.often, channelId: CHANNELS.helpJs }],
```
BTMPL, line 57: While I see the benefit for the jobBoard and helpReact channels I don't really see what benefit the helpJs and helpReact brings.

Maybe we should improve the new joiner experience in general? e.g. have a channel with a message explaining all of this visible right after joining and having other channels hidden for like 2 minutes, giving everyone time to read?

vcarl: I'd quite like to improve the new joiner experience in general, adding some form of onboarding would be awesome — right now we can't control what channel people enter, so I thought repeatedly surfacing info that that's tucked away in channel descriptions was a low-effort way to improve on the current state

ghardin137 (approved):

src/features/scheduled-messages.ts
nickmccurdy: We can always update the welcome screen channels (up to 5 supported).

PR #143, opened 2021-12-31
Use yarn, not npm

vcarl:
reactiflux/reactiflux.com already uses yarn, so it's been awkward for me as i switch back and forth between projects. I reach for yarn first, and since I'm the only one working on this full-time, I'm committing that to the repo


nickmccurdy:
We should use yarn 3 if we're going to switch, as yarn 1 is in legacy mode

nickmccurdy (changes requested):
We should use yarn 3 if we're going to switch, as yarn 1 is in legacy mode

vcarl:
The version installed by `npm i -g yarn` is 1.22, released Oct 21, 2021. yarn v1 seems fine to me, please justify this with a source beyond "it's in legacy mode" be https://github.com/yarnpkg/yarn/releases/tag/v1.22.17.

nickmccurdy:
Yarn support has moved to version 3 for features and bug fixes, see #144

PR #146, opened 2022-01-01
Small refactors

vcarl:
* Allow arrays of channel IDs/handlers in `addHandler`
* Rename `duration` -> `seconds` in `sleep()`, for clearer intelligence
* Use `CHANNELS.jobBoard` instead of string id

(This is a stacked diff off #145)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -141,6 +141,10 @@ addHandler("*", tsPlaygroundLinkShortener);
 
 bot.on("messageReactionAdd", handleReaction);
 
+bot.on("threadCreate", (thread) => {
+  thread.join();
+});
+
 bot.on("messageCreate", async (msg) => {
   if (msg.author?.id === bot.user?.id) return;
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,19 +79,23 @@ const channelHandlersById: ChannelHandlersById = {};
 
 const addHandler = (
   oneOrMoreChannels: string | string[],
-  channelHandlers: ChannelHandlers,
+  oneOrMoreHandlers: ChannelHandlers | ChannelHandlers[],
 ) => {
   const channels =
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -77,11 +78,27 @@ export type ChannelHandlersById = {
...
-const addHandler = (channelId: string, channelHandlers: ChannelHandlers) => {
-  channelHandlersById[channelId] = [
-    ...(channelHandlersById[channelId] || []),
-    channelHandlers,
-  ];
+const addHandler = (
+  oneOrMoreChannels: string | string[],
+  oneOrMoreHandlers: ChannelHandlers | ChannelHandlers[],
+) => {
+  const channels =
```
dylangarcia, line 85: `const channels = [].concat(oneOrMoreChannels);` is an easy way to do this btw

src/index.ts
```diff
@@ -77,11 +78,27 @@ export type ChannelHandlersById = {
...
+      : oneOrMoreChannels;
+  const handlers =
+    oneOrMoreHandlers instanceof Array
+      ? oneOrMoreHandlers
+      : [oneOrMoreHandlers];
+
+  channels.forEach((channelId) => {
+    const existingHandlers = channelHandlersById[channelId];
+    if (existingHandlers) {
+      existingHandlers.concat(handlers);
```
dylangarcia, line 97: This is a floating concat - `push` instead?

vcarl: whoop that's what it started as, good catch

src/index.ts
vcarl: TS complains, at least

dylangarcia: TS is stupid

nickmccurdy (approved):

Issue #134, opened 2021-12-18
Automatically create threads for specified channels


vcarl:
For specified channels:

* every message has a thread created for it, named something readable (name + date stamp?)
* replies are removed and the sender Dm'd instructions
* a !close command only the author can use (or maybe mvps and maybe a "helper" type role with this ability as well)
  * this is maybe not helpful, cuz afaik members can always close their own threads
* reactibot closes after 6 hours, and sends a message with instructions on following up
* instructions link to parent message and suggest replying to it if it's relevant

Padmanabh82:
I have more suggestions like, bot will pin the threads that are active or not being answered and thread's name should include timestamp, name of author, whole error or question (like why this unexpected behaviour is occurring)

vcarl:
Instead of a !close command, I think instead a ✅ reaction that will mark an answer as "accepted" and close the thread immediately is more helpful

nickmccurdy:
I believe we can also use buttons for this since we've upgraded the API.

PR #147, opened 2022-01-01
Add support for thread-only channels

vcarl:
This partially addresses the points in #134. It doesn't support timed closing, messaging to the thread on close, or adding a `!close` command, but it does enforce that threads must be used in a specific channel!

The rules:

* Top-level messages will have threads created with the name `name – time date`
* Replies to the message are deleted, with the messaged DM'd to the sender
* Replies to your own messages are _not_ removed (they stick around), intended to allow "followup" questions to be linked together

Video preview:

https://user-images.githubusercontent.com/1551487/147841572-bbd8362a-3ea8-45f3-a4d5-ea2c57c77db2.mov

(This is a stacked diff off #146)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,6 +4,7 @@ export const staffRoles = ["mvp", "moderator", "admin", "admins"];
 
 export const enum CHANNELS {
   "helpReact" = "103696749012467712",
+  "helpThreadsReact" = "565213527673929729",
   "helpJs" = "565213527673929729",
   "random" = "103325358643752960",
   "jobBoard" = "103882387330457600",
--- a/src/features/autothread.ts
+++ b/src/features/autothread.ts
@@ -1,0 +1,30 @@
+import { format } from "date-fns";
+import { sleep } from "../helpers/misc";
+import { ChannelHandlers } from "../types";
+
+const autoThread: ChannelHandlers = {
+  handleMessage: async ({ msg: maybeMessage }) => {
+    const msg = maybeMessage.partial
+      ? await maybeMessage.fetch()
+      : maybeMessage;
+
+    if (msg.type === "REPLY") {
+      const repliedTo = await msg.fetchReference();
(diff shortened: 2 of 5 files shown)


ghardin137 (approved):

vcarl:
From #134, this doesn't yet automatically close threads, and it doesn't provide a way for members to close their own threads.

I think I'm going to add a reaction to mark a question as answered (✅ or similar probably), and a scheduled task that scans all threads in the channel and closes any older than a specified threshold (probably 6 hours).

When a thread is closed as "resolved," reactibot will reply to the message (without pinging) with:

> This question has an answer! Thank you for helping 😄 
> 
> Threads are closed automatically after 6 hours, or if the member who started it reacts to a message with ✅. 

When a thread is closed for inactivity, reactibot will send:

> This thread has been closed automatically after 6 hours of inactivity. If you still need help, please ask a followup question in the main channel. If you think this thread has important context, please reply to your message with your new question!
> 
> {link to start of thread}

vcarl:
Now it does all those things!

Top-level view of a bunch of messages being archived:

https://user-images.githubusercontent.com/1551487/147863544-c591c0a5-c2c2-420d-8a0b-8b8e6068a086.mov

And the message that gets posted

tc pic archived message

This runs every half hour, and archives any threads that are older than 6 hours. One possible improvement: I had to add extra logic to detect empty threads, we could change the message to offer tips on asking a good question

Here's the entire experience, from thread opening to a response being marked as "accepted":

https://user-images.githubusercontent.com/1551487/147863554-025b1a24-7b45-4663-bbf6-17969609592d.mov

vcarl:
Also: 2 permission changes, one needed and one recommended.

* Disable "create thread" in thread channels, so all Top-level messages are questions (vs the "so and so created a thread" message)
* reactibot seems to need admin permissions in order to lock threads. I tried very hard to get it working without that, but that was the only thing that fixed it.

PR #148, opened 2022-01-02
Register slash commands

nickserv:


nickmccurdy:
I've pushed my slash command registration code if anyone would like to review or test it. For some reason the bot commands keep disappearing from Discord, so I'm still trying to work out some bugs.

vcarl:
@nickmccurdy what help testing do you need? I can smoke test, but I see there are some conflicts. Do you want to resolve or should I take a look? I think it's because of the `@everyone` consolidation, it's all under spam handling now

src/index.ts
```diff
@@ -54,6 +62,18 @@ bot
 
     if (bot.application) {
       const { id } = bot.application;
+
+      await api.put(Routes.applicationCommands(id), {
```
vcarl, line 66: Something else we may want to add, this is update-only. Might want to make sure we handle deleted commands too, by comparing `api.get(Routes.applicationCommands(id))` with the list of commands in the code

nickmccurdy:
I'm aware of the conflicts, I'll fix them.

I was referring to manually testing the slash commands on Discord. They seem to be pretty unpredictable. Specifically, if you try to list commands the Reactibot commands will rapidly flash on and off. I'm guessing it's either a Discord bug or undefined behavior in the API.

nickmccurdy:
The commands show up consistently when typing `/` now, but I'm not sure why Discord is claiming the command itself fails to send when the bot is actually receiving it and successfully responding.

image

PR #149, opened 2022-01-03
Cancel old builds

vcarl:
This should resolve an issue we just encountered where merging a second PR before the first build finished caused 2 instances of the bots to be left running. This is speculative, cuz actions are annoying to test

As 1 "build and run" step, I suspect that the cancellation won't have an opportunity to interrupt. By splitting it up, it increases the odds that we'll be able to abort on duplicate merges

Uses https://github.com/marketplace/actions/cancel-workflow-action

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -26,7 +26,32 @@ jobs:
         uses: actions/checkout@v1
         with:
           ref: ${{ github.ref }}
-      - name: Push to server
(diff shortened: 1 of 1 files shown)


dylangarcia (approved):

vcarl:
This looks to be broken, error from this step on multiple runs:
```
Error while canceling workflow_id 1668622: Cannot read property 'id' of null
```

Issue #123, opened 2021-12-05
!lock command


vcarl:
Something like, `!lock` to remove "send message" permissions for non-mods, useful in situations where there's a blowup and we need to figure it out quickly. A faster option than going into settings and toggling manually

taranvohra:
I'd like to work on this one if no one else is

PR #152, opened 2022-01-06
lock and unlock channel commands

taranvohra:
#123 

For `isStaffAdmin` utility, I'm assuming the member has `ADMINISTRATOR` permission checked in one of their roles (Admin, Moderator).

Also @vcarl can you add me to the test server. my discord user is **banny`#2169**

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -3,7 +3,7 @@ import fetch from "node-fetch";
 import { Message, TextChannel } from "discord.js";
 import cooldown from "./cooldown";
 import { ChannelHandlers } from "../types";
-import { isStaff } from "../helpers/discord";
+import { isStaff, isStaffAdmin } from "../helpers/discord";
 import { sleep } from "../helpers/misc";
 
 export const EMBED_COLOR = 7506394;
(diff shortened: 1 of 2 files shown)


src/features/commands.ts
```diff
@@ -669,6 +669,56 @@ Instead:
...
+      if (!msg.guild) {
+        return;
+      }
+
+      if (!isStaffAdmin(msg.member)) {
+        return;
+      }
+
+      // permission overwrites can only be applied on Guild Text Channels
+      if (msg.channel.type === "GUILD_TEXT") {
```
vcarl, line 686: Maybe:
```suggestion
      if (!msg.guild || !isStaffAdmin(msg.member)) {
        return;
      }

      // permission overwrites can only be applied on Guild Text Channels
      if (!msg.channel.isText()) {
```

vcarl: Oh, but if it needs to be a "guild text" maybe that doesn't work? Not sure if there's a similar check for DM/guild channel

vcarl:
@taranvohra Also, I think it's sufficient to use the existing `isStaff` check here. Moderators don't have administrator permissions, but I think we'd like them to be able to use this.

src/features/commands.ts
taranvohra: Yes,  it has to be narrowed to `guild_text` because `isText` implies it can also be `DM | Thread` and you cannot have permission overwrites on it

taranvohra:
@vcarl That'd also allow MVP roles to lock/unlock it. Would that be okay?

vcarl:
Aha, word. Sure, I think it's fine to extend to MVPs.

vcarl (approved):
Cool looks good!

PR #158, opened 2022-01-11
Give a short-lived instruction at the start of every thread

vcarl:
This posts an instructional message at the start of every thread, then deletes it after 30 seconds

> React to someone with ✅ to mark their response as the accepted answer and close this thread

### DIFF
--- a/src/features/autothread.ts
+++ b/src/features/autothread.ts
@@ -1,5 +1,6 @@
 import { differenceInHours, format } from "date-fns";
 import { Client, Message, PartialMessage } from "discord.js";
+import { CHANNELS } from "../constants";
 import { constructDiscordLink } from "../helpers/discord";
 import { sleep } from "../helpers/misc";
 import { ChannelHandlers } from "../types";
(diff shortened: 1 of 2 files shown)


nickmccurdy:
The code looks fine. Does it cause extra pings/notifications?

vcarl:
It doesn't! Opening the thread already provides a ping though

src/features/autothread.ts
```diff
@@ -60,9 +61,14 @@ const autoThread: ChannelHandlers = {
...
       }
     }
-    msg.startThread({
+    const newThread = await msg.startThread({
       name: `${msg.author.username} – ${format(new Date(), "HH-mm MMM d")}`,
     });
+    const message = await newThread.send(
+      `React to someone with ✅ to mark their response as the accepted answer and close this thread. If someone has been really helpful, give them a shoutout in <#${CHANNELS.thanks}>!`,
+    );
+    await sleep(30);
```
ScriptyChris, line 70: Maybe instead of arbitrary waiting for 30 seconds, it would be more UX friendly to show that instruction after somebody (a non-author) will post a message in the thread (or start counting 30 seconds after such post appears)? I think, ideally bot could recognize whether author replied with for example: "thanks", "it helped", "it worked" etc. and immediately notify the author about marking possibility, but that may require some difficult AI stuff. :)

nickmccurdy: Or we could simply send it after the first message in the thread

PR #162, opened 2022-01-17
Improve auto thread experience

vcarl:
* Emit metrics so we can track activity
* Allow members with helpful roles + staff to mark questions as answered
* Don't lock threads when answers are marked
* Share links to "how to ask good questions" resources if the thread was closed without any messages

### DIFF
--- a/src/features/autothread.ts
+++ b/src/features/autothread.ts
@@ -9,39 +9,6 @@ import { threadStats } from "../features/stats";
 const CHECKS = ["☑️", "✔️", "✅"];
 const IDLE_TIMEOUT = 12;
 
-const lockWithReply = ({
-  content: msg,
-  message,
-  starter,
-  shouldReply,
-}: {
-  content: string;
-  message: Message | PartialMessage;
-  starter: Message | PartialMessage;
-  shouldReply: boolean;
-}) => {
-  const { channel: thread } = message;
-  if (!thread.isThread()) {
-    return;
-  }
-
-  const content = `${msg}
-
-If you have a followup question, you may reply to this thread so other members know they're related. ${constructDiscordLink(
-    starter,
-  )}
-
-Threads are closed automatically after ${IDLE_TIMEOUT} hours, or if the member who started it reacts to a message with ✅ to mark that as the accepted answer.`;
-
-  (shouldReply
-    ? message.reply({ content, allowedMentions: { repliedUser: false } })
-    : thread.send(content)
-  ).then(() => {
-    thread.setLocked(true);
-    thread.setArchived(true);
-  });
-};
-
 const autoThread: ChannelHandlers = {
   handleMessage: async ({ msg: maybeMessage }) => {
     const msg = maybeMessage.partial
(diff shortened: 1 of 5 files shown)


src/features/autothread.ts
```diff
@@ -119,12 +116,24 @@ export const cleanupThreads = async (channelIds: string[], bot: Client) => {
...
+        thread
+          .send({
+            content: `This thread hasn’t had any activity in ${IDLE_TIMEOUT} hours, so it’s now locked.
+
+${
+  toCompare === starter
+    ? `Question not getting answered? Maybe it's hard to answer, or maybe you asked at a slow time. Check out these resources for help asking a good question:
+
+https://stackoverflow.com/help/how-to-ask
+http://wp.me/p2oIwo-26`
```
vcarl, line 129: New helper text if a thread is closed without any responses

src/features/stats.ts
```diff
@@ -32,19 +29,38 @@ const emitEvent = (
...
-  memberLeft: "member left server",
+const message = "message sent";
+const newMember = "new member joined";
+const memberLeft = "member left server";
+const threadCreated = "thread created";
+const threadReplyRemoved = "thread reply removed";
+const threadTimeout = "thread timeout";
+const threadResolved = "thread resolved";
+
+export const threadStats = {
```
ScriptyChris, line 40: Wouldn't it be safer to wrap this whole object in [`Object.freeze()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) (or mark it with [TypeScript's `as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions)) to prevent it from being modified (unless it is modified elsewhere) by any module, which imports it?

src/helpers/discord.ts
```diff
@@ -1,13 +1,39 @@
...
+import {
+  Message,
+  GuildMember,
+  PartialMessage,
+  Guild,
+  MessageReaction,
+  PartialMessageReaction,
+} from "discord.js";
+
+const staffRoles = ["mvp", "moderator", "admin", "admins"];
```
ScriptyChris, line 10: Same suggestion as [here](https://github.com/reactiflux/reactibot/pull/162/files#r786257564).

src/helpers/discord.ts
```diff
@@ -1,13 +1,39 @@
...
+const helpfulRoles = ["mvp", "star helper"];
```
ScriptyChris, line 11: Same as above.

Huuums:
Instead of just closing a thread when it gets marked as solved by the thread author we could try to implement some kind of timer mechanism. 

So when the thread gets marked as solved the bot posts a message like

> This question has been marked as solved. This thread will automatically close in 1 hour.

Once that message got posted people know they have about an hour to give more context to their answer if they want or the author asked for more clarification. 

We need to take a look how to actually implement this. Fetching the messages of a thread and searching for a message from reactibot might seem feasible for now but scaling could become an issue when there are many threads with lots of messages in the future.

ScriptyChris:
Nice. :)

I have one more feature suggestion (unless it already has been suggested): is it possible, to swap that `Jump` button (marked with red arrow on attached screenshot) for `Reply` or add them both? I don't know if it's really necessary inside the thread, but replying to one's post sends them ping (at least optionally), which is not the case with "normally" sent answer (unless replying user with explicitly added ping). It's like on some forums that have notifications system (StackOverflow for example) the OP gets notification if a answer appeared, which is convenient, because there is no need to "sit and wait" or periodically manually check the thread. If somebody doesn't want to be pinged then they might turn it off in channel/server/app settings.

image

src/features/stats.ts
vcarl: Probably, but I'm not too worried about it

vcarl:
> is it possible, to swap that `Jump` button (marked with red arrow on attached screenshot) for `Reply` or add them both?

Not that I know of, that's a Discord UI feature and we don't have control over it at all

PR #180, opened 2022-02-03
Fix "accept answer" double-sending in some cases

### LINKED ISSUE #173: Thread channels are double-sending "this has an answer" messages
The code to trigger a "this has an answer" bot response will fire every time someone else reacts, unintentionally sending multiple messages. We should be able to check who the current reaction is from, instead of whether the author has reacted, or we could read thread messages to see if the bot has already responded, or we could track a list of threads that have been "resolved".

definoob:
Check who the current reaction is from and reply only when the thread author reacted. Closes #173


src/features/autothread.ts
```diff
@@ -59,7 +59,7 @@ const autoThread: ChannelHandlers = {
       ? await thread.fetchStarterMessage()
       : undefined;
 
-    if (!starter || !guild) {
+    if (!starter || !guild || user.id !== starter.author.id) {
```
vcarl, line 62: This is actually desired behavior, we want to allow 2+ members who aren't the author to mark a response as the answer! see #162

src/features/autothread.ts
```diff
@@ -15,6 +16,20 @@ const CHECKS = ["☑️", "✔️", "✅"];
...
+): boolean => {
+  let flag = 0;
+
+  messages.forEach((message) => {
+    if (message.content.includes("This question has an answer!")) {
+      flag++;
+    }
+  });
+
+  return flag > 0;
```
vcarl, line 30: What about:

```suggestion
  return messages.some((message) =>
    message.content.includes("This question has an answer!")
  );
```

src/helpers/discord.ts
```diff
@@ -41,3 +42,6 @@ export const fetchReactionMembers = (
     .then((users) =>
       Promise.all(users.map((user) => guild.members.fetch(user.id))),
     );
+
+export const fetchMessagesByUser = (thread: ThreadChannel, userID: string) =>
+  thread.messages.cache.filter((message) => message.author.id === userID);
```
vcarl, line 47: This relies on the messages being cached, which won't be true for threads that were created before the bot restarted. Is it feasible to make this something like, `thread.messages.fetch()`? (I don't have a dev computer handy or I'd check, sorry)

src/features/autothread.ts
```diff
@@ -55,6 +70,16 @@ const autoThread: ChannelHandlers = {
     }
 
     const { channel: thread, author, guild } = await reaction.message.fetch();
+
+    const threadMessages = fetchMessagesByUser(
+      thread as ThreadChannel,
+      reactibot,
+    );
+
+    if (checkIfBotReplied(threadMessages)) {
```
vcarl, line 79: A couple notes:

* `fetchMessagesByUser` is a relatively expensive call, I think we should move it further down in the logic so it's only run when absolutely necessary -- we have less expensive calls that can bail out earlier
* I'd prefer to avoid `as`, as that bails out of type checking. There's a `thread.isThread()` assertion that perserves type safety. There's [an implied "return early if it's not a thread" here](https://github.com/reactiflux/reactibot/pull/180/files#diff-bdfc54759a662d98c57b52fe86e537b19ea177e8b8e2488a9a33723d2b74297bR83-R89), maybe we could make that more explicit and move this code further down to remove the need to cast

PR #182, opened 2022-02-04
!ban

Padmanabh82:
#175 !ban command completed

PR #183, opened 2022-02-04
Update constants

nickmccurdy:
I noticed a couple constants were out of date while reviewing #182


vcarl (approved):
Oh huh, I guess this has just always been working off the "mod-log" string match. Good catch

PR #184, opened 2022-02-05
added Code of Conduct command

intrvertmichael:
I find that it is a useful reminder when users are being inappropriate to remind them (or inform them) of the server's Code of Conduct. When doing so I usually have to open a new tab and search for it. It would be useful to have it as a command so that it can be easily accessible and more often remind users about them.


vcarl (approved):
Helpful addition!

PR #189, opened 2022-02-28
Add env detection to switch channel constants automatically

vcarl:
Also a couple of small fixes:

* Move stop/clean docker commands so there isn't a delay of few seconds where the bot isn't running
* Fix docker build to include `yarn.lock`
* Use `CHANNELS.botLog` instead of loading it from an environment variable


vcarl:
Follow up task here after merging: remove `BOT_LOG` secret from the repo

PR #195, opened 2022-03-21
Update !lift link to new docs

Aprillion:
"Lifting state up" was renamed in the new docs to "Sharing state between components" - https://beta.reactjs.org/learn/sharing-state-between-components - the spirit of the doc page is the same and I believe people here remember the `!lift` command, so I wouldn't change the command itself, just the link to the doc page.

Discussed in https://discord.com/channels/102860784329052160/955495969628233748


vcarl (approved):
Nice! Thanks for updating

PR #202, opened 2022-03-24
Actually fix command deploy?

vcarl:


vcarl:
Next time I'm using `feature/actions` so it runs in development

PR #205, opened 2022-03-26
Upload to GitHub Container Registry

vcarl:


.github/workflows/node.js.yml
```diff
@@ -9,12 +9,26 @@ jobs:
...
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Tag Build
+        uses: docker/metadata-action@v3
```
taranvohra, line 21: you need `id: meta` here

PR #206, opened 2022-03-28
Run CI build with Docker

vcarl:
Our "build" step is currently redundant, it installs/builds/tests, which allows the deployment step to install/build/run. If we uploaded docker images as a build artifact from this CI step, we could reuse it when deploying code.

I want to make that transition in several parts:

* (this PR) swap out "build" step to use docker and add ability to upload docker images (currently disabled as its unused)
* swap out deploy step to use the docker image built from CI
* (stretch) reduce custom deploy code and use more powerful build primitives. We're SSHing into a DO Droplet and running some commands directly on the server. DO exposes a managed Kubernetes product that looks pretty cheap and simple — if we could replace "ssh into a server and run docker" with "stand up a new k8s deployment," that would enable a lot more powerful work in the future

Major kudos to @taranvohra for getting a first draft of this working, and pointing out a bunch of dumb errors I made along the way.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,6 +1,6 @@
 name: Node.js CI
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -25,6 +25,8 @@ jobs:
           tags: |
             type=ref,event=pr
             type=ref,event=branch
+            type=sha
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
 
       - name: Build and push Docker images
         uses: docker/build-push-action@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/node.js.yml
```diff
@@ -9,12 +9,32 @@ jobs:
...
+            type=ref,event=pr
+            type=ref,event=branch
+            type=sha
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v2
+        with:
+          # push: ${{ github.event_name != 'pull_request' }}
+          push: false
```
vcarl, line 35: doesn't push currently

.github/workflows/node.js.yml
```diff
@@ -9,12 +9,32 @@ jobs:
...
+      - name: Tag Build
+        uses: docker/metadata-action@v3
+        id: meta
+        with:
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=ref,event=pr
+            type=ref,event=branch
+            type=sha
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
```
vcarl, line 29: Currently unused, but this is a PoC for future changes. It does in fact work! https://github.com/reactiflux/reactibot/pkgs/container/reactibot

PR #212, opened 2022-04-14
added code gif in features/commands.ts

sougataghar477:
Have a look :)

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -304,7 +304,7 @@ How To Ask Questions The Smart Way https://git.io/JKscV
 // short code snippets go here
 \\\`\\\`\\\`
 Like Shown Below 👇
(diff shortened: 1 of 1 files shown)


src/features/commands.ts
```diff
@@ -303,7 +303,8 @@ How To Ask Questions The Smart Way https://git.io/JKscV
             description: `\\\`\\\`\\\`js
 // short code snippets go here
 \\\`\\\`\\\`
-
+Like Shown Below 👇
+https://im5.ezgif.com/tmp/ezgif-5-26189abd64.gif
```
vcarl, line 307: This is 404ing for me, and I'm not sure how comfortable I am linking out to a random domain like this. Is there a way to get it as en embed on the message?

vcarl:
Dupe of #282

PR #215, opened 2022-04-17
Thread/Replies for warnings in job-board

AldrinMathew:
- The warnings posted by the bot in the job-board channel is now sent in
  a thread. If the thread creation fails the same warning is sent as a
  reply to the message in question

### DIFF
--- a/src/features/jobs-moderation.ts
+++ b/src/features/jobs-moderation.ts
@@ -165,19 +165,26 @@ const jobModeration = async (bot: Client) => {
       if (cooldown.hasCooldown(message.author.id, "user.jobs")) return;
 
       cooldown.addCooldown(message.author.id, "user.jobs");
-      message.author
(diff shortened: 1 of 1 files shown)


src/features/jobs-moderation.ts
```diff
@@ -165,19 +165,26 @@ const jobModeration = async (bot: Client) => {
...
-[HIRING] - you are looking to hire someone
-[INTERN] - this is an intern position, no experience required
-[REMOTE] - only remote work is possible
-[LOCAL] - only local work is possible (please remember to provide the country / city!)
-[VISA] - Your company will help with the visa process in case of successful hire
-
-Thank you :)
-
-:robot: This message was sent by a bot, please do not respond to it - in case of additional questions / issues, please contact one of our mods!`);
+      const thread = await message.startThread({ name: 'Warning' });
```
vcarl, line 168: This is more of a suggestion than a warning, since we use "warning" to mean moderator interventions elsewhere I don't want to confuse
```suggestion
      const thread = await message.startThread({ name: message.author.username });
```

vcarl:
Thanks for doing this!

PR #223, opened 2022-05-23
Escape mod log pings and links for real, lower spam threshold 4 > 3

vcarl:


src/features/autodelete-spam.ts
```diff
@@ -20,6 +21,37 @@ const safeDomains = [
 const checkWords = (message: string, wordList: string[]) =>
   message.split(/\b/).some((word) => wordList.includes(word.toLowerCase()));
 
+const getPingCount = (content: string) => {
```
vcarl, line 24: I wanted to use [`message.pings.everyone`](https://discord.js.org/#/docs/discord.js/stable/class/MessageMentions?scrollTo=everyone) but it isn't true if the ping didn't work, apparently. Have to parse them out of the text

PR #224, opened 2022-05-24
Use npm instead of yarn

vcarl:
I realized that Yarn is reliably including the cache in the final docker image, which explains why final size was larger than I expected. This is quite a bit smaller

```
$ docker images
REPOSITORY        SIZE
<none>            245MB
vcarl/reactibot   288MB
```

I'm grumpy because I prefer yarn haha. But can't argue with 15% smaller artifacts


vcarl:
Yeah huge difference on published artifacts
```
$ docker images
REPOSITORY                    TAG          SIZE
ghcr.io/reactiflux/reactibot  sha-2227036  246MB
ghcr.io/reactiflux/reactibot  main         302MB
```

nickmccurdy:
Could we disable Yarn's cache or move it somewhere else?

vcarl:
Not without other downsides. I spent several days trying

PR #229, opened 2022-06-06
Never ping users or roles mentioned in scheduled messages

vcarl:


vcarl:
I poked around to see if I could change it so that the Wordle ping wouldn't send a notification, but doing so means that those with the role don't get added to the thread. womb womp

PR #236, opened 2022-06-29
Add initial !const command

awareness481:
Based on a recent discussion in #typescript I think it's safe to say that there's still confusion around how const works and it not being immutable.
Let me know if you want this so I can work on it

Edit: Preview of the embed on my machine

image

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -734,7 +734,7 @@ Instead:
   },
   {
     words: [`!const`],
-    help: `explains that const variables are not immutable`,
+    help: `explains that const values are not immutable`,
     category: "Web",
     handleMessage: (msg) => {
       msg.channel.send({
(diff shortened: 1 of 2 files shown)


awareness481:
For some reason I can't access the token of the bots I create through the web app, so I can only view these changes when I get on my home computer instead.
I ll leave this as draft for now

awareness481:
I accidentally commited the package-lock changes, should probably be included in the PR?

src/features/commands.ts
```diff
@@ -734,16 +734,36 @@ Instead:
...
+> \`\`\`
+or you could mutate an array's elements
+> \`\`\`js
+const arr = ['hello'];
+arr.push('world'); // ✅
+arr[1] = 'reactiflux'; // ✅
+console.log(arr); // ['hello', 'reactiflux']
+> \`\`\`
+
+https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#:~:text=The%20const%20declaration%20creates%20a%20read%2Donly%20reference%20to%20a%20value.%20It%20does%20not%20mean%20the%20value%20it%20holds%20is%20immutable%E2%80%94just%20that%20the%20variable%20identifier%20cannot%20be%20reassigned.%20For%20instance%2C%20in%20the%20case%20where%20the%20content%20is%20an%20object%2C%20this%20means%20the%20object%27s%20contents%20(e.g.%2C%20its%20properties)%20can%20be%20altered.
```
vcarl: the `#:~:text=` thing only works in Chrome (and it's huuuge, let's replace with a link to the section?
```suggestion
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#description
```

nickmccurdy:
Do you think it would be helpful to mention `Object.freeze` and `Readonly`, or would that be too wordy?

src/features/commands.ts
awareness481: Done, but I also changed it from the md link format `[]()` to just the link, because md links will open a popup where the user has to hit confirm to continue.

I havent tested it locally yet (cant find where the flip I have the repo of reactibot with the env vars :P ) but imo it should be fine.

Didnt know the text links only worked in chrome, thought it was some cool new browser standard :(

vcarl (approved):

vcarl:
TY! I think this is about max length for an embed, we could add another link to a resource if we want to highlight something else specific

PR #240, opened 2022-07-25
Optionally ping user within bot command's message

ScriptyChris:
Hey, i noticed that it's a bit inconvenient that often people have to ping a folk, which they just notified about some hint (like how to properly attach a code or how to properly ask a question) via the bot - so first message is calling the bot with i.e. `!code` command and, after the bot replies, one has to add like "hey @ScriptyChris please take a look :point_up: " to actually notify other person about the given hint. So i prepared an enhancement that if a certain command is used with pings added and/or it's a reply, then bot will automatically forward these pings in a form of prefixed message to the response command.

Currently (for starters) the feature is hooked up with commands: `!ask`, `!code (!gist)`, `!ping`.
Looking forward for a feedback. :+1: 

I hope that below screens present the change more readably than my description. :)

**Before:**

image

**After:**

image
image

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -16,8 +16,12 @@ const prepareOptionalPingForward = ({ users, roles, repliedUser }: MessageMentio
     pingedReceivers.unshift(formatReceiverID(repliedUser.id));
   }
 
+  if (!pingedReceivers.length) {
+    return {};
+  }
+
   return {
-    content: `Hey ${pingedReceivers.join(', ')}! :wave: Please check the hint below. :point_down:`
+    content: `Hey ${pingedReceivers.join(', ')}! :wave: Please check the hint below. :point_down: \n\u200B`
   };
 }
 
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -1,11 +1,25 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
 import fetch from "node-fetch";
-import { Client, Message, TextChannel } from "discord.js";
+import { Client, Message, TextChannel, MessageMentions } from "discord.js";
 import cooldown from "./cooldown";
 import { ChannelHandlers } from "../types";
 import { isStaff } from "../helpers/discord";
 
 import * as report from "../commands/report";
+
+const prepareOptionalPingForward = ({ users, roles, repliedUser }: MessageMentions) => {
+  const formatReceiverID = (receiverID: string) => `<@${receiverID}>`;
+  const pingedReceiversIDs = [...users.keys(), ...roles.keys()];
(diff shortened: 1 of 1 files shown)


nickmccurdy:
This could also be a good use case for application commands (which we're already using for moderation)

vcarl:
I'm not sure that there's much confusion over who the intended recipient is when these commands are invoked, is that something you've noticed much of?

ScriptyChris:
@nickmccurdy What application commands do you mean?
@vcarl I think that some people (especially new comers) might not be aware that a bot's hint is directed to them. I guess this is why they are often pinged below the bot's reply - this enhancement is just making it a bit more convenient.

vcarl:
My concern is that this change results in those members getting pinged twice, which I think is confusing in a different direction. It's not obviously a straight improvement to the experience in my eyes

nickmccurdy:
> @nickmccurdy What application commands do you mean?

@ScriptyChris I meant to say message commands. They're a type of application command that shows up when you open the context menu for a message.

> My concern is that this change results in those members getting pinged twice, which I think is confusing in a different direction. It's not obviously a straight improvement to the experience in my eyes

@vcarl That's a good point. Could we fix this by having the mention in a slash command?

ScriptyChris:
@nickmccurdy you mean this?

image

What would be the difference between invoking a command (i.e. `!code`) as it is now vs from that menu?

@vcarl you mean "pinged twice", because some might not notice that the bot already pinged a person? It's fair argument, although if somebody would ping a person while using the command, then i don't think they would do it again after bot posts a message. Also, this enhancement is opt-in, because if one doesn't suffix the command with ping, then bot won't use that ping, so it will work as it has been so far. Also, i don't see a much of an issue if somebody would be pinged in two messages in a row - like... it can already happen if two different people reply to a person using the ping.

vcarl:
How does this look, functionally? Would need some other supporting logic to help transition from `!commands` to slash commands but i wanted to rough out an approach

Screen Shot 2022-07-27 at 10 18 05 AM
Screen Shot 2022-07-27 at 10 18 14 AM
Screen Shot 2022-07-27 at 10 18 31 AM


https://user-images.githubusercontent.com/1551487/181272498-e155b3bf-28e7-40c4-b16e-0d86165dd63a.mp4

ScriptyChris:
To me it looks more powerful than my enhancement proposal, because multiple commands can be used at once and there is command name auto-completion (as opposed to `!<command-name>` syntax). So :+1: from me.

nickmccurdy:
Can you open a PR so I can also push? I'd like to try to merge my WIP into this. That should help us support commands like MDN with more args.

PR #241, opened 2022-07-28
Use slash commands instead of message listeners for stored replies

### LINKED ISSUE #131: Enable application commands
[Application commands](https://discord.com/developers/docs/interactions/application-commands) provide a better UX for running bot commands in Discord while also allowing for more relaxed bot permissions.

vcarl:
Alternative approach to #148, will close #131 

This puts every `!command` under a single top-level `/r` command, with an autocompleted list of replies as the first argument. It includes the idea introduced by #240 (thanks @ScriptyChris!) of telling the bot who to ping — #240 had the problem of double-pinging, but as a slash command argument it can be done without that drawback.

This code removes some commands for simplicity, it's a functional proof of concept more than final code. Remove commands:

* `!mdn`
* `!lock`
* `!unlock`

Another change: `!commands` has been changed to `/r help`, which uses an ephemeral reply to avoid shoving everything off scroll back.

This is very much a rough draft, functional but incomplete. Before merging we should add logic to encourage folks to shift from `!commands` syntax to slash commands and figure out how to handle `!mdn`

Demo:

Screen Shot 2022-07-27 at 10 18 14 AM

Screen Shot 2022-07-27 at 10 18 31 AM

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -1,11 +1,7 @@
-/* eslint-disable @typescript-eslint/no-use-before-define */
-import fetch from "node-fetch";
-import { Client, Message, TextChannel } from "discord.js";
-import cooldown from "./cooldown";
-import { ChannelHandlers } from "../types";
-import { isStaff } from "../helpers/discord";
+import { Client, InteractionReplyOptions } from "discord.js";
 
 import * as report from "../commands/report";
+import { SlashCommandBuilder } from "@discordjs/builders";
 
 export const setupInteractions = (bot: Client) => {
   bot.on("interactionCreate", (interaction) => {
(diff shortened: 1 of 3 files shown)


src/features/commands.ts
```diff
@@ -310,197 +290,143 @@ Link a Code Sandbox to an existing GitHub repo: https://codesandbox.io/s/github/
...
-        ],
-      });
-
-      fetchMsg.delete();
+          color: EMBED_COLOR,
+        },
+      ],
     },
   },
+  // {
```
ScriptyChris, line 315: Shouldn't this long commented section be removed (i guess it's a remnant from code testing)?

vcarl: Yeah this is more of a draft state atm, roughed out. I'll update description to add necessary followup before this can merge

PR #243, opened 2022-07-29
Updating the mdn url to match the new api structure to fix the mdn bot

gristoi:
MDN  have changed the structure of their api, moving the locale from being a url parameter to being a query parameter.
Have tested on private server with reactibot


vcarl (approved):

vcarl:
Nice! ty!

PR #253, opened 2022-09-20
Feat: Adds remote work command to bot

AveryTheDev:
### Overview

A lot of folks have been asking for help with acquiring remote jobs over the years, especially recently. This PR adds a command to the bot that allows a user to use either `!remote` or `!remotework` to view links related to this subject.

Given the relatively high number of junior developers asking about remote work, I've included a note section warning folks at this stage of their career that acquiring a remote job is incredibly rare. I have opted to do so given that this is generally stated by many who reply to this question on the Reactiflux server to optimize the amount of time saved.

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -770,7 +770,7 @@ https://exploringjs.com/es6/ch_variables.html#_pitfall-const-does-not-make-the-v
     },
   },
   {
-    words: [!remote, !remotework],
+    words: ["!remote", "!remotework"],
     help: "provides resources for the remote work job search",
     category: "Web",
     handleMessage: (msg) => {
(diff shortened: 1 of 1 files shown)


src/features/commands.ts
```diff
@@ -769,6 +769,39 @@ https://exploringjs.com/es6/ch_variables.html#_pitfall-const-does-not-make-the-v
       });
     },
   },
+  {
+    words: [!remote, !remotework],
```
vcarl, line 773:
```suggestion
    words: ["!remote", "!remotework"],
```

AveryTheDev: Sorry about that. It must've somehow changed from a string to vars when I tried to copy and paste this blurb I wrote out elsewhere. Fixed!

vcarl (approved):
Whoohoo, great addition! tyty

Issue #262, opened 2022-10-26
create a command for react keys


CaptainTux:
the topic seems to come up quite frequently, either explicitly or implicitly
I think it would be useful to have a `!keys` command for such situations.
for example today I saw someone using UUIDs as keys which got generated on every render, and it wasn't the first time
perhaps something like (I'm bad with coming up with educational messages, forgive me, there should probably be another sentence or 2)

> Keys are important if you map over an array and the callback returns JSX
> *some more here*
> For further information see
> - https://beta.reactjs.org/learn/rendering-lists#keeping-list-items-in-order-with-key
> - https://kentcdodds.com/blog/understanding-reacts-key-prop

louis-young:
I think this is a great suggestion. Thanks for suggesting it and raising an issue @CaptainTux. 

I've tweaked the blurb slightly and raised https://github.com/reactiflux/reactibot/pull/263.

PR #263, opened 2022-10-29
Feature/#262: Add command to explain the importance of keys when rendering lists in React

louis-young:
PR for #262

This pull request adds a `!keys` command that explains the importance of keys when rendering lists in React. It also references two authoritative resources with further information.

### Changes

- Add `!keys` command.

### Screenshots

image

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -847,7 +847,9 @@ https://remotive.io/remote-jobs/software-dev
             title: "The importance of keys when rendering lists in React",
             type: "rich",
             description: `
-React depends on the use of stable and unique keys to identify items in a list so that it can perform correct and performant DOM updates. Keys are particularly important if the list can change over time.
+React depends on the use of stable and unique keys to identify items in a list so that it can perform correct and performant DOM updates. 
+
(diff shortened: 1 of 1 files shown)


mattvanvoorst-contentful:
LGTM tbh, wondering if you maybe need to say something about it being okay to use `index` as a key for non-changing lists? Unless that's covered in the resources.

louis-young:
> LGTM tbh, wondering if you maybe need to say something about it being okay to use `index` as a key for non-changing lists? Unless that's covered in the resources.

Thanks Matt 🙌 

Good shout, I think it’s worth clarifying. Do you think the note about the importance of keys in lists that change over time could be extended to communicate explicitly that they are less important in lists that don’t change and that React will use the index by default if no key is provided, or do you think it warrants its own paragraph?

mattvanvoorst-contentful:
Maybe a new paragraph with a single line: `You might not always have a unique key available. If you're sure your list will not change it's fine to use map's index as key`

louis-young:
> Maybe a new paragraph with a single line: `You might not always have a unique key available. If you're sure your list will not change it's fine to use map's index as key`

I've updated the command 😄 

image

The latest macOS update has bricked my machine so I'll sort that and then commit.

mattvanvoorst-contentful (approved):
LGTM

PR #264, opened 2022-11-10
chore(commands): update state update command links

bpas247:
Removes a link to an old post to be updated with a more reliable source (old post will most likely be deleted soon).


louis-young (approved):
It's nice to see this covered explicitly in the new documentation. Thanks for providing the original gist and for raising the pull request 🎉

Issue #271, opened 2023-01-10
Implement feature to remove low-effort questions and encourage the author to post a more reasonable question


louis-young:
Recently, the help channels have been barraged with low-effort questions. This contains but is not limited to:

1. Error messages, stack traces, screenshots, and general questions lacking crucial context/information/code.
2. Statements like "Help it does not work.".
3. Questions such as "Can anyone help me?".

I think this has led to a decline in the number of active helpful users in the help channels as channels that are primarily comprised of a sea of such questions are not inviting to helpful contributors. It also encourages others to post similar questions by normalising such questions. I think we should strive to achieve a higher average quality of questions.

I think we should educate the authors and encourage them to ask more reasonable questions.

We propose that we should implement a bot feature that given a number of reactions from members (including one authoritative reaction from a rank such as a star helper or moderator) deletes the message and messages the author, providing educative information on how to ask for programming help and encourages them to ask a more reasonable/answerable question.

Thanks to @mattmattvanvoorst for suggesting this.

PR #272, opened 2023-01-10
Feature/#271: Write low effort question remover feature

louis-young:
PR for #271 

This pull request implements the bot feature that removes low-effort questions, educates the authors, and encourages them to ask more reasonable questions.

The feature:

- Creates a new thread from the offending message.
- Sends an educative and encouraging message to the thread to help the user improve their question (thanks @vcarl!).
- Deletes the offending message to clean up the channel.
- Logs the deletion in the moderation log.

### Changes

- Create and consume a shared utility function to create new thread names (I did this before changing the thread name/format subsequently but decided to leave it in).
- Extend the emoji moderation feature to include a reaction handler for low-effort questions.
- Extend the report log to support low-effort question removal logs.

### Notes

- ~I'm leaving the blurb up for debate. What message do we want to send to the author of the message?~
- I felt like it was sensible to extend the emoji moderation feature as essentially this is a form of emoji moderation, although I'm happy to separate this out should anyone strongly disagree. 
- I co-located the constants/configuration for this feature with the handler as it seemed sensible. Let me know if you disagree and would prefer it lifted up to module scope. 
- This introduces some conflation between helpers and staff in the moderation log as we're considering helpers to be authoritative here so that they can "approve" a message to be deleted. This means it will include helpers in the staff list in the report in the moderation log. I didn't think it was worth making the changes to make this distinction clearer but let me know if you disagree.
- To clarify behaviour, this expects at least one helpful or staff member to react _and_ at least two total reactions (so one helper, one member/one staff, one member/one staff one helper) to trigger.

### DIFF
--- a/src/features/autothread.ts
+++ b/src/features/autothread.ts
@@ -10,6 +10,7 @@ import {
 import { sleep } from "../helpers/misc";
 import { ChannelHandlers } from "../types";
 import { threadStats } from "../features/stats";
+import { createNewThreadName } from "../helpers/threads";
 
 const CHECKS = ["☑️", "✔️", "✅"];
 const IDLE_TIMEOUT = 12;
@@ -39,7 +40,7 @@ const autoThread: ChannelHandlers = {
     }
     // Create threads
     const newThread = await msg.startThread({
-      name: `${msg.author.username} – ${format(new Date(), "HH-mm MMM d")}`,
+      name: createNewThreadName({ username: msg.author.username }),
     });
     threadStats.threadCreated(msg.channelId);
     // Send short-lived instructions
--- a/src/features/emojiMod.ts
+++ b/src/features/emojiMod.ts
@@ -5,6 +5,8 @@ import { ReportReasons, reportUser } from "../helpers/modLog";
 import { fetchReactionMembers, isHelpful, isStaff } from "../helpers/discord";
 import { partition } from "../helpers/array";
 import { sleep } from "../helpers/misc";
+import { EMBED_COLOR } from "./commands";
+import { createNewThreadName } from "../helpers/threads";
 
 const config = {
   // This is how many ️️warning reactions a post must get until it's considered an official warning
@@ -83,9 +85,28 @@ export const reactionHandlers: ReactionHandlers = {
       return;
     }
 
-    await message.delete();
+    const newThread = await message.startThread({
+      name: createNewThreadName({ username: message.author.username }),
+    });
 
-    await message.author.send("...");
+    await newThread.send({
+      embeds: [
+        {
+          title: "Please improve your question",
+          type: "rich",
+          description: `
+            Sorry, our most active helpers have flagged this as a question that needs more work before a good answer can be given. This may be because it's ambiguous, too broad, or otherwise challenging to answer. 
+
+            This is a good resource about asking good programming questions: 
+
+            https://zellwk.com/blog/asking-questions/
+          `,
+          color: EMBED_COLOR,
+        },
+      ],
+    });
+
+    await message.delete();
   },
 };
 
--- a/src/features/emojiMod.ts
+++ b/src/features/emojiMod.ts
@@ -119,8 +119,8 @@ https://zellwk.com/blog/asking-questions/
     reportUser({
       reason: ReportReasons.lowEffortQuestionRemoved,
       message,
-      staff: unknownReactors,
-      members: knownReactors,
+      staff: knownReactors,
+      members: unknownReactors,
     });
   },
 };
--- a/src/features/emojiMod.ts
+++ b/src/features/emojiMod.ts
@@ -2,7 +2,7 @@ import { MessageReaction, Message, GuildMember, Guild } from "discord.js";
 import cooldown from "./cooldown";
 import { ChannelHandlers } from "../types";
 import { ReportReasons, reportUser } from "../helpers/modLog";
-import { fetchReactionMembers, isStaff } from "../helpers/discord";
+import { fetchReactionMembers, isHelpful, isStaff } from "../helpers/discord";
 import { partition } from "../helpers/array";
 import { sleep } from "../helpers/misc";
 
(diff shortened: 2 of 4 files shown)


src/features/emojiMod.ts
```diff
@@ -69,6 +69,24 @@ export const reactionHandlers: ReactionHandlers = {
...
+      (userWhoReacted) => isStaff(userWhoReacted) || isHelpful(userWhoReacted),
+    );
+
+    if (!hasRequiredNumberOfReactions || !hasAuthoritativeUserReacted) {
+      return;
+    }
+
+    await message.delete();
+
+    await message.author.send("...");
```
louis-young, line 88: I'm leaving the blurb up for debate. What message do we want to send to the author of the message?

vcarl: Do you think it would be feasible to set this up to reply in a thread, instead of DMing the author? I think DMs are easier to miss, and more people have them disabled (which would mean the author gets no indication for why their message was removed). That adds more complexity, like what happens if the message is already in a thread, but I think it's a better experience

louis-young: I agree with the downsides of using a direct message 👍 perhaps sending the message in the channel and mentioning the author would solve this and also act as a deterrent and provide passive information to other users? I think a thread is likely feasible if you feel strongly about it, although I'm keen on removing the messages to really try and strongly enforce a higher quality of question 🙂

vcarl: For a blurb, maybe something like
> Sorry, our most active helpers have flagged this as a question that needs more work before a good answer can be given. This may be because it's ambiguous, too broad, or otherwise challenging to answer. This is a good resource about asking good programming questions: https://zellwk.com/blog/asking-questions/

louis-young: Absolutely perfect, thanks. I'll use that 🎉

vcarl: > I think a thread is likely feasible if you feel strongly about it, although I'm keen on removing the messages to really try and strongly enforce a higher quality of question 🙂
Agreed! Actually I was thinking of threads as an ideal way to approach that, in fact — can open a thread, send a message to it, then remove the initial message, leaving very little trace of both the question and the intervention to an outside observer

louis-young: If this behaves as you expect, I'll guard against reactions within a thread and add the reporting functionality 🎉

src/features/autothread.ts
```diff
@@ -39,7 +40,7 @@ const autoThread: ChannelHandlers = {
     }
     // Create threads
     const newThread = await msg.startThread({
-      name: `${msg.author.username} – ${format(new Date(), "HH-mm MMM d")}`,
+      name: createNewThreadName({ username: msg.author.username }),
```
mattvanvoorst-contentful, line 43: Should we make this thread name more generic and recognisable as something that was marked as a poor question? Right now it's just the authors name, maybe we could consider something along the lines of:

`Sorry ${msg.author.username}, your question needs more work`

src/features/emojiMod.ts
```diff
@@ -69,6 +71,43 @@ export const reactionHandlers: ReactionHandlers = {
...
+    });
+
+    await newThread.send({
+      embeds: [
+        {
+          title: "Please improve your question",
+          type: "rich",
+          description: `
+            Sorry, our most active helpers have flagged this as a question that needs more work before a good answer can be given. This may be because it's ambiguous, too broad, or otherwise challenging to answer. 
+
```
mattvanvoorst-contentful, line 99: Could we consider adding a line that we prefer questions with !code attached, instead of screenshots or photos of code?

src/features/autothread.ts
louis-young: Yeah good idea, it's not specific enough. Your suggestion looks good to me 👍

mattvanvoorst-contentful (approved):
I approve of this message

src/features/emojiMod.ts
louis-young: Yeah, I think that'd be worthwhile 👍 do you have a suggestion or shall we take an excerpt from the !code command?

mattvanvoorst-contentful: Maybe have the bot send another message with the !code command? Double down on the spam?

louis-young: I'm not actually sure if the bot can trigger itself, I'd have to have a look at that. I'll try it and go from there

vcarl: Maybe it would be good to link to a set of Discord tips? Maybe a new page on reactiflux.com, perhaps?

louis-young: Yeah, that's a nice idea 👍  perhaps we can look at doing that in a second pass as I'd like to release this feature as soon as possible. What do you think?

louis-young:
@vcarl and @mattvanvoorst-contentful, I've finished this feature. I have:

- Improved the thread name as per Matt's suggestion.
- Added protection against reactions to messages in threads.
- Logged the low-effort question deletion in the moderation log.

I've updated the description and written descriptive commits. Please let me know what you think. I'm keen to roll out this feature 🙂

src/features/emojiMod.ts
louis-young: I've done this 👍

src/features/emojiMod.ts
louis-young: The bot can't trigger itself and I'm not sure it makes sense to start pulling the command outputs out so they can be reused at this point. I like Carl's suggestion and I'd be happy to release this as-is and iterate on it in the future when we've had time to build a new page on the website. What do you think?

vcarl: Into deferring a new page for late 👍

src/features/emojiMod.ts
```diff
@@ -69,6 +70,59 @@ export const reactionHandlers: ReactionHandlers = {
...
+      ],
+    });
+
+    await message.delete();
+
+    reportUser({
+      reason: ReportReasons.lowEffortQuestionRemoved,
+      message,
+      staff: unknownReactors,
+      members: knownReactors,
```
louis-young, line 123: I think we need to switch these around. Aside from that it looks good to go 🎉 

```suggestion
      staff: knownReactors,
      members: unknownReactors,
```

louis-young:
Thanks everyone, this looks good to go 🎉

PR #275, opened 2023-01-19
Update variable names for brevity

vcarl:
I also added a parenthetical explaining how the message got posted, to make it easier to know the feature exists

### DIFF
--- a/src/features/emojiMod.ts
+++ b/src/features/emojiMod.ts
@@ -71,29 +71,20 @@ export const reactionHandlers: ReactionHandlers = {
     reportUser({ reason, message, staff, members });
   },
   "🔍": async ({ message, usersWhoReacted }) => {
-    const NUMBER_OF_AUTHORITATIVE_REACTIONS_REQUIRED_FOR_DELETION = 1;
-    const NUMBER_OF_REACTIONS_REQUIRED_FOR_DELETION = 2;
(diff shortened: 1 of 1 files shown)


src/features/emojiMod.ts
```diff
@@ -71,22 +71,19 @@ export const reactionHandlers: ReactionHandlers = {
...
+    const KNOWN_REACTOR_THRESHOLD = 1;
+    const UNKNOWN_REACTOR_THRESHOLD = 2;
 
-    const [unauthoritativeUsersWhoReacted, authoritativeUsersWhoReacted] =
-      partition(
-        (userWhoReacted) =>
-          isStaff(userWhoReacted) || isHelpful(userWhoReacted),
-        usersWhoReacted,
-      );
+    const [knownReactors, unknownReactors] = partition(
```
louis-young, line 77: Are these the wrong way around? I think it partitions items that match the predicate into the second (or first if we're zero-based) array and I'd expect the helpful/staff partition to be named as known and the regular users as unknown but perhaps I'm looking at it wrong (apologies if so).

louis-young (commented):
The name changes look great 🎉  just one question about the partition

src/features/emojiMod.ts
vcarl: You're right! Fixed

PR #296, opened 2023-07-05
(TS 5.1) Job board – Validation errors, observability, cleanup

vcarl:
- Refactors for code structure, types
- Add a bunch more observability. It should be much easier to keep an eye on this now
- Fix a longstanding scheduler bug
- Fix false positive "too frequent" when editing messages
- Fix validating every private thread message, including outside the job board
- Fix a bunch of borked validation types. Unfortunately the types are brittle.
- Improve rendering of validation error messages, and provide more information back to the poster about length
- Fixes for enforcement thread and forehire post cleanup. Not sure these fixes work, because I had manually tested it previously. Would like to steal some mocking code out of [AnswerOverflow](https://github.com/AnswerOverflow/AnswerOverflow), Rhys has mentioned a lot of cool work there


vcarl:
TS 5.1 update broke types, so I reverted that. I updated to get a feature used in 4.9, but didn't end up committing that code

PR #315, opened 2023-09-14
Upgrade Node for security

nickmccurdy:
[Node 16 is dead](https://nodejs.org/en/blog/announcements/nodejs16-eol) and we need to upgrade for support and security (OpenSSL is no longer supported). In the past we seemed to use the latest Node version, so I'm upgrading to that for now.

I don't have Docker locally but everything seems to be fine on Node 20.6.1.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -18,7 +18,10 @@ jobs:
         uses: actions/checkout@v2
 
       # TODO: see if we need to do anything special to cache docker images
-      - uses: c-hive/gha-npm-cache@v1
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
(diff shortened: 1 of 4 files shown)


vcarl:
Have you been able to test that this works? I'm wary of touching the Actions unfortunately because of how annoying they are to test. Same to reactiflux/reactiflux.com#273 and and reactiflux/mod-bot#37

nickmccurdy:
Yes:
> I don't have Docker locally but everything seems to be fine on Node 20.6.1.

vcarl:
I'm not concerned about Node 20 or the Actions platform, but afaik the `Deployment` action hasn't been run yet, looking at action history. Have you positively confirmed that this action produces a running instance of Reactibot?

PR #335, opened 2023-11-27
Adjust jobs moderation bot

vcarl:
* Require links in HIRING posts
* Double the lifespan of a FORHIRE post to 60 hours
* Don't delete edited messages. If we hit problems with formatting evasion, we'll need to reevaluate. Edited posts still get flagged, just not deleted.
* Small copy adjustments on validation failure thread creation
* Small adjustments to support test environment


vcarl:
Hack fix for #321 in that it doesn't delete them anymore, but it's not resolved.

Issue #337, opened 2023-12-06
Add !frameworks command


barrybtw:
**!frameworks**

A command to display all the popular frameworks, maybe even show different results depending on the channel? 

_help_react_
- next.js
- remix
- etc

_help_backend_
- express
- koa
- etc

Or as 2 seperate commands? I feel like 1 command to show all becomes a big chunk where a lot of items might be irrelevant to a discussion.

PR #338, opened 2023-12-06
Feature/frameworks command #337

barrybtw:
**Feature**
`!frameworks/!framework` command to display an embed of all the popular frontend, backend, meta & non-js http frameworks that exist.

image

**Issue**
#337

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 yarn.lock
 *.log
 k8s-context
+pnpm-lock.yaml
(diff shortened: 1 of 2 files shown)


vcarl (approved):
Cheers, nice! I feel vaguely like "frameworks" here might be a muddy term — express or React I think might be appropriately called "libraries" or just tools, but understand why it makes sense to include them on a list like this. I'm going to bias in favor of merging vs nitpicking language tho, can iterate if necessary.

vcarl (approved):

Issue #339, opened 2023-12-06
Add !ui command


barrybtw:
!ui

A command to display all the popular UI libraries & styling tools.

vcarl:
Closed in #340

PR #340, opened 2023-12-06
Feature/UI command

barrybtw:
**Feature**
`!ui` command to display an embed of popular UI libraries & tools.

**Actual embed shown**
Embed image

**Issue**
#339


pnpm-lock.yaml
vcarl, line 1: Can you push a commit to remove this? 👀

Issue #348, opened 2024-01-08
Add message command to search React documentation


kristersd:
Star helpers frequently reference React documentation to address commonly asked questions. Adding a Reactibot command integrated with the documentation could quickly direct users to relevant sections for resolving their issues. For instance, using the command `!react createContext` would provide a brief explanation and direct the user to the appropriate documentation section.

**Implementation Ideas**
- "Scraping" the `react.dev` source code from GitHub would allow Reactibot to display `<Intro />` from their MDX, along with appropriate documentation links. This extraction can be achieved using simple regular expressions. Link to their [GitHub](https://github.com/reactjs/react.dev/tree/main/src/content/reference).
  - Example of their **[MDX](https://raw.githubusercontent.com/reactjs/react.dev/main/src/content/reference/react/useRef.md)**.

**Challenges**
- In instances where `<Intro />` contains references or links to other parts of the documentation, these links should be correctly updated to maintain their functionality.
- Not always the command would point to `react` directory and sometimes, we may need to reference `react-dom` API. 
  -  Star helpers are often aware which API they are referencing, it could be different command, e.g `!react-dom createPortal`

**Questions**
- Does GitHub have any limitations on the number of requests that can be made and would we surpass them? We may need to consider caching and saving documentation on each new Reactibot version.
  - There is limited subset of developers, which would choose to use the command.

nickmccurdy:
Sounds like an interesting idea! We already have something similar with `!mdn`, though we reference the React docs in enough commands that this would probably save us more time. I hope we wouldn't need much more infra to consume the React docs, though.

> Does GitHub have any limitations on the number of requests that can be made and would we surpass them? 

Yes, there are separate rate limits for [REST](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api) and [GraphQL](https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api). We already have a GitHub token in our environment though, so I don't think we'd run into issues with this.

PR #349, opened 2024-01-18
Implement search React documentation command #348

kristersd:
Implemented additional commands `!docs` and `!react-docs` to search React documentation. The output embed will contain a link to documentation and short intro segment from their page.

Extra:
- Fixed MDN `icon_url` not being found.

reactibot

### DIFF
--- a/src/helpers/react-docs.ts
+++ b/src/helpers/react-docs.ts
@@ -30,17 +30,19 @@ export const getReactDocsContent = async (searchPath: string) => {
 };
 
 export const getReactDocsSearchKey = (search: string) => {
-  const searchKey = REACT_AVAILABLE_DOCS.find((key) => {
-    const matches = key.split("/");
-    const name = matches[matches.length - 1];
-    const namespace = matches.length <= 2 ? key : `${matches[0]}/${matches[2]}`;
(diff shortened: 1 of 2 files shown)


src/helpers/react-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  if (matches.length > 0) {
+    const [introContent] = matches.map(([, match]) => match.trim());
+    return introContent;
+  }
+
+  return null;
+};
+
+const REACT_AVAILABLE_DOCS = [
```
kristersd, line 63: I am not quite sure about having constant defined here and not schedule task top update it, but the react documentation updates to file path are not frequent and this may become out of date only after major releases.

Do we want to have additional requests to populate this array or having an generate script would be enough, when necessary? What do you think?

vcarl: Yeah, I think this is a reasonable assumption. If we find it breaks more frequently than is convenient then we can reevaluate, but given effort to repair breakage + likely frequency of breakage, this seems reasonable

vcarl:
This is maybe a candidate for swapping out to a slash command + autocomplete, I think there's prior art for that in #241

PR #353, opened 2024-02-07
Fix docs access, the token it relied on expired quickly after deploy

vcarl:


kristersd (approved):

kristersd:
Don't forget to update the cluster/workflow with the appropriate key name. May rename the `env.example` as well.

vcarl:
Oh good point, cheers. Heckin deployment fiddliness

PR #355, opened 2024-03-10
fix: react documentation commands can not be in sentence #354

### LINKED ISSUE #354: Search React documentation doesn't recognize keywords correctly
While the message contains more than `!docs useState` e.g `For example, !docs useState` the search documentation will not be generated correctly.  We should only filter out the query after `!docs/!react-docs`

kristersd:
Closes #354

### DIFF
--- a/src/helpers/react-docs.test.ts
+++ b/src/helpers/react-docs.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, test } from "vitest";
+import { extractSearchKey, getReactDocsSearchKey } from "./react-docs";
+
+describe("React documentation command", () => {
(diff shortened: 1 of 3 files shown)


src/helpers/react-docs.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, test } from "vitest";
+import { extractSearchKey, getReactDocsSearchKey } from "./react-docs";
+
+describe("React documentation command", () => {
```
vcarl, line 4: :tada:

vcarl (approved):
Looks great! tyty

PR #356, opened 2024-03-10
Add pdf generation and resume review resources #320

### LINKED ISSUE #320: Post recommended resources for anyone who posts to #resume-review

kristersd:
Added ability to generate png images from pdf files posted in #resume-review and some suggestions.
I have pinned the package versions and created [issue](https://github.com/reactiflux/reactibot/issues/357) to update the packages as introducing pdf2pic caused the build to fail.

- Removed `discord-api-types` in package.json as they seem to not be used anywhere

Closes #320.


src/index.ts
```diff
@@ -120,7 +121,10 @@ const handleMessage = async (message: Message) => {
   }
   const msg = message.partial ? await message.fetch() : message;
 
-  const channelId = msg.channel.id;
+  const channel = msg.channel;
+  const channelId = channel.isThread()
```
kristersd, line 125: I tested the other message handlers and they seemed to work, but this could add them as well in threads and do we want that? This could be a breaking change

vcarl: This seems fine to me, treating threads as if they came in via the parent channel makes intuitive sense to me

vcarl (approved):

PR #360, opened 2024-03-19
Add early return to thread enforcement

vcarl:


vcarl:
I think what might have happened is, thread messages started coming through on more parts of the API? We started getting a new error message on the 11th, but it's not super clear how this is related to the change in behavior

image


Okay yeah this looks to have been incorrectly running through all the "should a thread be created" logic on messages sent in threads. It blew up on startThread because it's "not in a guild text channel" (it's in a thread)

So it would leave regular messages alone, but deleted replies because they're processed earlier and have an early return

It's possible that "onMessage" wasn't receiving all thread messages before or something, but it's a little weird that it ever worked correctly tbh. Seems like it should always have had this behavior 🤔

PR #367, opened 2024-05-07
reactibot-modlogs-and-debug-events

mattmattvanvoorst:
Added a custom bot command to render a set of debug buttons, also drafted up a first set of mod events


vcarl (approved):
Couldn't test debug command locally but code looks good! Only adding it in non-prod env makes sense to me, that makes it low-risk enough that I'm comfortable merging

src/features/mod-activity.ts
```diff
@@ -0,0 +1,111 @@
...
+
+const guildMemberTimeoutHandler = (
+  oldMember: GuildMember | PartialGuildMember,
+  newMember: GuildMember,
+) => {
+  const oldTimeout = oldMember.communicationDisabledUntil;
+  const newTimeout = newMember.communicationDisabledUntil;
+
+  if (!oldTimeout && newTimeout) {
+    logger.log(
```
vcarl, line 17: Oh – these may make more sense to put into #mod-log, but the existing helpers for doing so are a little heavier-weight than just logging. It's got different types of logs defined as constants and a function to render them appropriately, which might not make sense in this context

mattmattvanvoorst: Sure thing! The previous PR enables just that - they now report to the `modLog`

PR #375, opened 2024-06-04
Auto thread title fix, Resume Review features

### LINKED ISSUE #373: AI resume review revisions
There have been a couple model updates that are probably worth investigating, and we need to make this more visible by automatically adding it to the screenshot post

While we're in here, we should look at making it all send as 1 message, as a smol UX tweak. "1 message" on every thread seems better than 2 to me

### LINKED ISSUE #374: Thread creation fails to pull Twitter info
It falls back to the discord name of the poster, which sucks. Maybe a problem with the x.com domain

vcarl:
- Fixes #374 
- Fixes #373 
- Adds a "delete" button to #resume-review posts so that post authors can clear out their private info


vcarl:
Resume Review AI demo:

https://github.com/reactiflux/reactibot/assets/1551487/3250b720-f8ad-4c1d-a6cd-e32c0b36268f

src/index.ts
```diff
@@ -191,8 +189,6 @@ addHandler(
 const threadChannels = [CHANNELS.helpJs, CHANNELS.helpThreadsReact];
 addHandler(threadChannels, autothread);
 
-addHandler(CHANNELS.resumeReview, resumeReviewPdf);
```
kristersd, line 194: We are tying the image generation to the `threadCreate` and when user has not uploaded PDF in their thread and has it done in the occurring message it will not generate the image.

I think we should leave the cooldown and this implementation in-place. Perhaps there is way to differentiate when user has created thread or posted a message in already created thread.

kristersd (changes requested):

src/index.ts
vcarl: Gotcha, fair point on behavior change! I'll revisit

vcarl: Pushed changes!

vcarl: https://github.com/reactiflux/reactibot/assets/1551487/65078e44-ab06-4718-a233-d6d691fd6f76

src/features/resume-review.ts
```diff
@@ -25,34 +36,55 @@ const sendResumeMessage = async (msg: Message): Promise<true | void> => {
...
+      );
       return;
     }
 
-    const sucess = await sendResumeMessage(message);
-    if (sucess) {
+    await message.channel.sendTyping();
+
+    const images = await buildResumeImages(message);
+    if (images.length > 0) {
```
kristersd, line 65:
```suggestion
    if (images.length <= 0) return;
```

We should do early-return here to not send embed to the user without any images.

src/features/resume-review.ts
```diff
@@ -25,34 +36,55 @@ const sendResumeMessage = async (msg: Message): Promise<true | void> => {
...
+      !message.channel.isThread() ||
+      message.channel.parentId !== CHANNELS.resumeReview
+    ) {
+      return;
+    }
+    const cooldownKey = `resume-${message.channelId}`;
 
     if (cooldown.hasCooldown(message.author.id, cooldownKey)) {
+      message.channel.send(
+        "You posted < 5 minutes ago, not creating a new preview",
```
kristersd, line 57:
```suggestion
        "You posted just a few minutes ago. Please wait a bit before creating a new preview.",
```

Let's make it a bit friendlier 😄

kristersd (approved):

src/features/resume-review.ts
vcarl: :grumpy:

src/features/resume-review.ts
vcarl: Oh word yeah. This suggestion is gonna bork the close brace but good point

src/features/resume-review.ts
```diff
@@ -25,34 +36,55 @@ const sendResumeMessage = async (msg: Message): Promise<true | void> => {
...
+    await message.channel.sendTyping();
+
+    const images = await buildResumeImages(message);
+    if (images.length > 0) {
       cooldown.addCooldown(
         message.author.id,
         cooldownKey,
         FIVE_MINUTES_IN_SECONDS,
       );
     }
```
vcarl, line 71:
```suggestion
    if (images.length === 0) {
      return;
    }

    cooldown.addCooldown(
      message.author.id,
      cooldownKey,
      FIVE_MINUTES_IN_SECONDS,
    );
```

src/features/resume-review.ts
vcarl: See https://github.com/reactiflux/reactibot/pull/375/files/211b595e840bbf8c9c363ceafb6bf5d1632174fd#r1629821110

PR #381, opened 2024-06-08
#357 deps: upgrade dependencies and Typescript versions

### LINKED ISSUE #357: Upgrade packages and Typescript version
We have recently pinned the versions in [merge requeust](https://github.com/reactiflux/reactibot/pull/356) as adding new package has failed the build. We should update the packages and Typescript version.

kristersd:
- Updated dependencies and Typescript version.

Some of the libraries e.g `querystring`, `node-fetch` are not updated as it would require to use dynamic import. We should probably just move away from `commonjs` module.

Closes #357

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "reactibot",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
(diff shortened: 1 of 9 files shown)


vcarl (approved):
Heck yeah, thanks for taking care of the casts and keeping this up-to-date

PR #387, opened 2024-06-12
fix: this context lost on destruct

kristersd:
Fixes issue when `this` is `undefined` when we have destructed the `fetchStarterMessage`.

```js
[ERROR] Cannot read properties of undefined (reading 'parent') TypeError: Cannot read properties of undefined (reading 'parent')
    at fetchStarterMessage (/build/reactibot/node_modules/discord.js/src/structures/ThreadChannel.js:315:26)
```

### DIFF
--- a/src/features/resume.ts
+++ b/src/features/resume.ts
@@ -76,8 +76,12 @@ export const resumeResources = async (bot: Client) => {
       deferred.edit("Looking for a resume…");
       const messages = await interaction.channel.messages.fetch();
 
-      const { fetchStarterMessage } = interaction.channel;
-      const firstMessage = await retry(() => fetchStarterMessage(), 5, 10);
+      const channel = interaction.channel;
+      const firstMessage = await retry(
+        () => channel.fetchStarterMessage(),
+        5,
+        10,
+      );
       if (!firstMessage) {
         await interaction.reply({
           ephemeral: true,


src/features/resume.ts
```diff
@@ -76,8 +76,12 @@ export const resumeResources = async (bot: Client) => {
       deferred.edit("Looking for a resume…");
       const messages = await interaction.channel.messages.fetch();
 
-      const { fetchStarterMessage } = interaction.channel;
-      const firstMessage = await retry(() => fetchStarterMessage(), 5, 10);
+      const channel = interaction.channel;
+      const firstMessage = await retry(
+        () => channel.fetchStarterMessage(),
```
matt-miro, line 81: If we couldn't deconstruct `channel` from `interaction.channel` before because it can be falsy, you might want to nullcheck for `channel` here just to be sure?

kristersd: The underlying issue is not about the value being `null` other than when you have implementation relying on `this` it will lose the context when destructed. For example,
```js
class Foo {
  parent = ""
  
  getParent() {
    return this.parent;
  }
}

const foo = new Foo();
const { getParent } = foo;
getParent() // Uncaught TypeError: Cannot read properties of undefined (reading 'parent')
```

mattmattvanvoorst (approved):
Shippit

vcarl (approved):
Hecc this is dumb. Totally forgot (never knew?) that scope could be effected in this way

PR #395, opened 2024-08-06
fix: add error logging and retry to threadCreate resume

kristersd:
Add additional error logging for the resume feature and handle retrying the firstMessage when thread has been created

### DIFF
--- a/src/features/resume.ts
+++ b/src/features/resume.ts
@@ -4,6 +4,7 @@ import {
   InteractionType,
   ComponentType,
   ChannelType,
+  Message,
 } from "discord.js";
 import OpenAI from "openai";
 import { CHANNELS } from "../constants/channels";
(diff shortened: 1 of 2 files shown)


src/features/resume.ts
```diff
@@ -79,8 +79,11 @@ export const resumeResources = async (bot: Client) => {
       const channel = interaction.channel;
       const firstMessage = await retry(
         () => channel.fetchStarterMessage(),
-        5,
-        10,
+        () => logger.log("RESUME", "Failed to fetch interaction first message"),
```
vcarl, line 82: What about doing this with a catch block instead of modifying the helper 👀 Thoughts?

src/features/resume.ts
```diff
@@ -207,7 +210,18 @@ export const resumeResources = async (bot: Client) => {
     if (thread.parentId !== CHANNELS.resumeReview) {
       return;
     }
-    const firstMessage = await thread.fetchStarterMessage();
+
+    const firstMessage = await retry(
+      () => thread.fetchStarterMessage(),
+      () => {
+        logger.log("RESUME", "Failed to fetch thread first message");
+      },
```
vcarl, line 218: samesies

vcarl (approved):

src/features/resume.ts
kristersd: I had implemented it with try/catch but we kind of ended up with this scenario, do you feel like this would be better? I would prefer to not mix await with `await retry().catch(void 0)`, if that was your initial idea.
```ts
let firstMessage: Message<true> | null = null;
try {
  firstMessage = await retry(void 0)
} catch {
  logger.log(void 0),
}
if (!firstMessage) {
  void 0
}
```

vcarl: I was thinking probably to put a wider try{} block but yea error handling ergonomics suck for promises. This invites other `onX` changes, generally I think it's good to avoid introducing new APIs to helper functions without a strong justification. For me personally this didn't rise to it but not something I'm gonna block a PR over either.

PR #396, opened 2024-08-06
Don’t do PR previews, which resolves #216

### LINKED ISSUE #216: External collaborator PRs fail CI build
See [this run](https://github.com/reactiflux/reactibot/runs/6101864939?check_suite_focus=true) for instance:

```
#18 ERROR: denied: installation not allowed to Write organization package
------
 > pushing ghcr.io/reactiflux/reactibot:pr-215 with docker:
------
error: denied: installation not allowed to Write organization package
Error: buildx failed with: error: denied: installation not allowed to Write organization package
```

vcarl:


vcarl:
Fixes #216 (I thought the commit should be enough but it isn't?)

Issue #398, opened 2024-08-10
feature: !auth command to explain popular libraries for authentication


barrybtw:
Would be nice to have a command to share information about authentication as to not repeat ourselves daily. I'll get on it.

barrybtw:
pr added, #399 solves this issue

PR #399, opened 2024-08-10
feature: added command !auth solves #398

barrybtw:
Added command `!auth` to show authentication resources, additionally you can add a user to the reply as such and the user will be apart of the embed:
image

**Example reply from bot**
image

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -1152,7 +1152,9 @@ _ _
         title: "",
         type: EmbedType.Rich,
         description: `
-Authentication is a critical part of most web applications. Below are some resources to help you get started with authentication:
+Authentication is a critical part of most web applications. Here are some resources to help you get started.
+- [JSON Web Tokens (JWT) are Dangerous for User Sessions](https://redis.io/blog/json-web-tokens-jwt-are-dangerous-for-user-sessions/)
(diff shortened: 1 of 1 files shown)


src/features/commands.ts
```diff
@@ -1141,6 +1141,60 @@ _ _
...
+    help: "Provides a list of popular backend- and meta-frameworks",
+    category: "Web",
+    handleMessage: (msg) => {
+      const firstMention = msg.mentions.users.first();
+
+      const embed = {
+        title: "",
+        type: EmbedType.Rich,
+        description: `
+Authentication is a critical part of most web applications. Below are some resources to help you get started with authentication:
```
vcarl, line 1155: I'd love to also mention something about JWTs here, whatcha think? Putting it here vs in the 'fields' in part because titles are longer, and to call it out as more opinion than resource

```suggestion
Authentication is a critical part of most web applications. Here are some resources to help you get started.
- [JSON Web Tokens (JWT) are Dangerous for User Sessions](https://redis.io/blog/json-web-tokens-jwt-are-dangerous-for-user-sessions/)
- [JWT should not be your default for sessions](https://evertpot.com/jwt-is-a-bad-default/)
```

barrybtw: Agree 👍🏼

vcarl (approved):

PR #409, opened 2024-08-22
Remove reddit auto-post

vcarl:
This is getting moved into its own repo with a different execution environment (https://github.com/reactiflux/feeds, running on vcarl's home computer)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,12 +14,11 @@
         "discord.js": "14.15.3",
         "dotenv": "16.4.5",
         "gists": "2.0.0",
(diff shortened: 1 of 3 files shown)


vcarl:
Reddit blocks API access from datacenter IPs @ashwat07, so this works locally but not when deployed to DigitalOcean. I tried applying for an API key but never heard back, so simplest fix was to move to another repo and deploy it to a 24/7 computer in my living room 🤣

vcarl:
And they're old.reddit.com URLs because I'm a stubborn old bastard who doesn't like new reddit :grumpy: hahaha

Issue #393, opened 2024-07-30
Expose #job-board via an API


vcarl:
This interacts with the reactiflux.com repo, messages posted that don't meet our validation requirements are not removed from the /jobs page. This is very much not ideal and limits our usefulness as a job space, because spammy and duplicate postings remain.

We should replace the current system with a reflection of the currently-active posts in the server. We monitor as far back as 90 days, with skill ads (`forhire` posts) kept up for several days.

I'm thinking the simplest way to do this would be to set up a Cloudflare Worker with a KV store to track posts. Since we scan the channel for posts on startup, the lack of persistence isn't a problem. Then, on the website, we can read the list of posts and present a similar experience. Long-term, it would be nice to improve the experience, and maybe consider presenting a list of skill ads that are currently active.

vcarl:
At first I thought we'd need to set up a database for this, but actually I guess not? It's all stored in memory, scanning history for active posts. Maybe it'll make sense to parse further back into history on startup, so we can give more than a single week's backlog, but I don't know that a full-ass database is the way to go there.

vcarl:
Shipped in #439

PR #427, opened 2025-01-02
Add an API for active job posts

vcarl:
This might need some tweaking but I'm pushing up a solid first effort

Screenshot 2025-01-01 at 8 36 19 PM

I'll need to ship some DNS changes before this merges


vcarl:
Resolves #393, will require followup work in for https://github.com/reactiflux/reactiflux.com/issues/321

PR #450, opened 2025-03-02
Exclude code blocks when looking for !command triggers - Fixes #412

### LINKED ISSUE #412: text in code blocks is treated as commands
Text written in code blocks is treated as commands. We should ignore code blocks, for example, to avoid triggering the bot when the user has:
```js
if (!auth) {}
```

**Implementation details:**
- Ignore text fragments which are written in code blocks.

achyu-dev:
I have tried to fix the issue #412  that is text in code block as comments. 

The code I have used to fix this is Attached below - 

```

// Helper functions to handle commands inside codeblocks
export function removeCodeBlocksAndQuotes(text: string): string {
  if (!text) return '';
  let filtered = text.replace(/```[\s\S]*?```/g, '');
  filtered = filtered.replace(/`[^`]+`/g, '');
  filtered = filtered.replace(/^>.*$/gm, '');
  filtered = filtered.replace(/>>>[\s\S]*?(?=\n\n|$)/g, '');
  filtered = filtered.replace(/\n{3,}/g, '\n\n').trim();
  
  return filtered;
}

export function isInsideCodeBlock(text: string, position: number): boolean {
  // Count backticks before position
  const beforeText = text.slice(0, position);
  const singleBackticks = (beforeText.match(/`/g) || []).length;
  const tripleBackticks = (beforeText.match(/```/g) || []).length;
  
  // Inside inline code block if odd number of single backticks
  if (singleBackticks % 2 !== 0) return true;
  
  // Inside multi-line code block if odd number of triple backticks
  if (tripleBackticks % 2 !== 0) return true;
  
  return false;
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "type": "module",
   "scripts": {
     "start": "node dist/index.js",
-    "dev": "npm-run-all --parallel 'dev:*'",
+    "dev": "npm-run-all --parallel dev:bot dev:test",
     "dev:bot": "tsx --watch ./src/index.ts",
     "dev:test": "vitest",
     "build": "tsc -b",
@@ -45,6 +45,7 @@
     "xss": "^1.0.15"
   },
   "devDependencies": {
+    "@types/jest": "^29.5.14",
     "@types/node": "20.14.2",
     "@types/node-cron": "3.0.11",
     "@types/node-fetch": "2.6.11",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -53,13 +54,14 @@
     "@typescript-eslint/parser": "7.12.0",
     "eslint": "8.57.0",
     "eslint-config-prettier": "8.8.0",
+    "jest": "^29.7.0",
```
vcarl, line 57: Can you uninstall this? Blocker for merge

{
vcarl, line 1: Please delete 🙏

src/features/emojiMod.ts
```diff
@@ -23,11 +24,8 @@ import { partition } from "../helpers/array.js";
 import { EMBED_COLOR } from "./commands.js";
 
 const config = {
-  // This is how many ️️warning reactions a post must get until it's considered an official warning
```
vcarl, line 26: Hm why delete these? Self-documenting?

src/features/commands.ts
```diff
@@ -1268,7 +1301,8 @@ const commands: ChannelHandlers = {
 
     commandsList.forEach((command) => {
       const keyword = command.words.find((word) => {
-        return msg.content.toLowerCase().includes(word);
+        //return msg.content.toLowerCase().includes(word);
```
vcarl, line 1304: :eyes: delete plz

package.json
achyu-dev: sure will do

PR #461, opened 2025-06-11
feat(experimental): added a job/introduction bot detection

Simplyalex99:
**description**:  

- Added a bot scanner to remove messages of job postings in channels:  help-react, help-js, help-react-native, help-styling, help-backend, general-react,general-tech. After 5 minutes, the reply message from the bot disappears.
- Added missing channels help-react-native, help-styling, help-backend, general-react,general-tech to channels.ts.

I have attached a video as a preview of this new feature. For some context, the first two messages in the video are about a job posting, which gets removed by the bot, and the last message is legit and does not get removed.

https://github.com/user-attachments/assets/f0ebbc9a-2391-4430-aae0-e6fa016bdda5

***update***: The bot now solely removes posts related to job postings rather than both intro and job posts to simplify the process and prevent false positives/negatives. Added an edge case for code blocks with dollar signs inside for languages like jQuery and other libraries. Any attempt to bypass the bot by including code blocks with dollar a sign and other job keywords gets marked as a job posting ensuring consistency and accuracy.


Simplyalex99:
**update**: After careful inspection, the .includes function is returning false positives for legitimate post like "help me". Currently working on using regex instead of .includes.

Simplyalex99:
**update**: Fixed incorrect messages being marked as a job posting with partial job keywords rather than the full job keywords by using regex to test for full keywords in job keywords. Removed currencies from job keywords into a separate array to test if a message contains any substring of currencies. 

Also added two more keywords that 100% should be marked as a job posting such as "we are hiring". The code is ready to be merged.

PR #466, opened 2025-06-15
feat: Job board moderation bot detects insecure link

### LINKED ISSUE #456: Job board format moderation doesn't seem to detect insecure links
Looks like we're not identifying `http://` urls as links, which probably shouldn't be the case

Simplyalex99:
**description**: This pull-request resolves #456 

Below is a preview of the results:
bot-preview

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -74,6 +74,23 @@ const commandsList: Command[] = [
     },
   },
   {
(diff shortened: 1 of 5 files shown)


Simplyalex99:
Plan on reopening PR after isolating the cross post commit.

PR #469, opened 2025-06-16
update: added option in regex for insecure urls

### LINKED ISSUE #456: Job board format moderation doesn't seem to detect insecure links

Simplyalex99:
This pull-request resolves #456 by adding a special "?" to accept insecure or secure URLs.


vcarl:
:seemsgood:

This resolves #456 but I will call out [discussion](https://github.com/reactiflux/reactibot/issues/456#issuecomment-2985752317) that it may be preferable to just not allow insecure links generally 🤷

PR #471, opened 2025-06-18
feat(experimental): added job keywords detection [resubmitted after fix]

Simplyalex99:
This pull request is a fixed version of #461 . 

**description**:  

- Added a bot scanner to remove messages of job postings in channels:  help-react, help-js, help-react-native, help-styling, help-backend, general-react,general-tech. After 5 minutes, the reply message from the bot disappears.
- Added missing channels help-react-native, help-styling, help-backend, general-react,general-tech to channels.ts.

I have attached a video as a preview of this new feature. For some context, the first two messages in the video are about a job posting, which gets removed by the bot, and the last message is legit and does not get removed.

https://github.com/user-attachments/assets/f0ebbc9a-2391-4430-aae0-e6fa016bdda5

***update***: The bot now solely removes posts related to job postings rather than both intro and job posts to simplify the process and prevent false positives/negatives. Added an edge case for code blocks with dollar signs inside for languages like jQuery and other libraries. Any attempt to bypass the bot by including code blocks with dollar a sign and other job keywords gets marked as a job posting ensuring consistency and accuracy.


src/features/commands.ts
```diff
@@ -73,40 +73,6 @@ const commandsList: Command[] = [
       });
     },
   },
-  {
-    words: [`!contribution`],
```
vcarl, line 77: 👀 intentional?

Simplyalex99: 👀  no, good catch not sure how I missed that even after checking. Will fix it asap

PR #486, opened 2025-07-05
fix: undefined properties and removed colon in print statement

Simplyalex99:
### Summary
Switched from fetchStarterMessage to just fetching the first message in the thread and checking for any attachments or content before sending a bot response. In addition, added a retry in case of race conditions. Also removed colon from print statement for better readability.

### Issue
The error Cannot read properties of undefined (reading 'parent') originated from fetchStarterMessage when a forum post only contains an image and no description. 

### Solution
However, after care analysis, the fetchStarterMessage seems to only apply for non-forum channels as forums are different and do not have a parent id post, unlike normal threads when tested in development. For this reason, the fetchStarterMessage was removed for now.

### DIFF
--- a/src/features/looking-for-group.ts
+++ b/src/features/looking-for-group.ts
@@ -50,7 +50,18 @@ export const lookingForGroup = async (bot: Client) => {
     }
 
     try {
-      await retry(thread.fetchStarterMessage);
+      const messages = await retry(() => thread.messages.fetch({ limit: 1 }));
+      const message = messages.first();
+      if (!message) {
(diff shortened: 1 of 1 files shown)


Simplyalex99:
Updated! And apologies, I couldn't replicate the issue on production earlier but now I am seeing it for every post on bot log. The issue reproduction was a case that caused this to happen on both local and production when I first tried it.

Simplyalex99:
After testing, it does appear to be a context issue of losing this without passing a callback.

Issue #498, opened 2025-08-22
Lucia deprecated as a library, replace with better-auth


barrybtw:
As lucia-auth has been deprecated as a library and is now deemed a learning resource, it could perhaps be moved to learning resources under the `!auth` command.

As lucia docs state:
> Lucia v3 has been deprecated. Lucia is now a learning resource for implementing sessions and more.

An addition to the libraries row could be https://www.better-auth.com/

PR #499, opened 2025-08-22
Lucia Moved to learning resource

achyu-dev:
This PR solves #498  issue and thus moves Lucia to a learning resource and adds better auth to the libraries


barrybtw (changes requested):
I would personally want the list to go from _easiest to implement -> hardest to implement_, i find better-auth to be way easier to use and navigate than Passport, i don't know how you feel about it. So i propose authjs, better-auth, passport

PR #522, opened 2026-08-26
chore: remove xCancel embed

asbedb:
Removed the xCancel functionality as it is currently not functional.

This feature served the community well and provided an alternative route to view and participate in discourse without barriers.

~ Thank you xCancel.

A note from xCancel below:

> On Monday 24th August at 8PM EST, we received at letter from X Corp. asking to cease and desist the service XCancel.
The service XCancel is stopped until further notice. We are seeking legal advice and won't share more details for now. Thank you for the trust you have put in these two years of XCancel.


asbedb:
@vcarl  - over to you (friendly reminder I can't merge 😂 ).

